PR #1136, opened 2013-07-15
Disk caching of `UIImageView+AFNetworking` category

robertmryan:
I have attempted to add disk caching of `UIImageView+AFNetworking` category. I've made best efforts to ensure backward compatibility (i.e. it defaults to RAM-only cache unless you invoke the variation of the methods with an `options` parameter in which you specify disk-based caching as well).

In an otherwise quite simple change, the one wrinkle was that `AFImageRequestOperation` did not expose the `responseData`, so I had to make new methods in `AFImageRequestOperation`. As a result, although I would have liked to limit my edits to the `UIImageView` category, I had to add a couple of methods to `AFImageRequestOperation`.

The one change that I'm particularly apprehensive about is my removal of `setCompletionBlockWithSuccess`. I removed this because I needed it to return the `NSData` for the `responseObject`, and as that's the standard completion block behavior, I concluded that `setCompletionBlockWithSuccess` was no longer necessary. If you really need it, I can restore the `setCompletionBlockWithSuccess` that has the `UIImage` variation of `responseObject`. But I'll have to implement some other rendition that returns the `NSData` object rather than the `UIImage` object.

By the way, while I did my `AFImageRequestOperation` modifications (to add a variation that supports returning the `NSData` as well) to optimizing disk caching, this might be a useful feature in general. When retrieving an image in the network, the naive intuition, to use `UIImagePNGRepresentation` or `UIImageJPEGRepresentation` is often not a good idea. You lose meta data, you can either lose image quality or make the file larger (if original was already JPEG) or both, etc. It's really quite important when retrieving images to be able to capture the original resource, the `NSData`, rather than only getting the `UIImage`. For lots of web stuff, sure, the `UIImage` is fine, but for serious image stuff, having access to the original `NSData` can be very useful.

Anyway, there you go. There's no pride of authorship here, so I'm open to comments, suggestions and other feedback. But disk caching of asynchronous image requests is a very useful feature. Currently when someone on Stack Overflow needs disk-based caching (for example, [this is an example of a recent question](http://stackoverflow.com/questions/17653663/)), we have to tell them, "er, well, then don't use AFNetworking, but use something like SDWebImage instead." And for me, at least, it's rare that I want caching of images, and _don't_ want disk caching, and I can't be the only person for whom this is the case. So disk caching of images seems like a very logical extension.

### DIFF
--- a/AFNetworking/AFImageRequestOperation.m
+++ b/AFNetworking/AFImageRequestOperation.m
@@ -354,39 +354,4 @@ static UIImage * AFInflatedImageFromResponseWithDataAtScale(NSHTTPURLResponse *r
     return [_acceptablePathExtension containsObject:[[request URL] pathExtension]] || [super canProcessRequest:request];
 }
 
-//- (void)setCompletionBlockWithSuccess:(void (^)(AFHTTPRequestOperation *operation, id responseObject))success
-//                              failure:(void (^)(AFHTTPRequestOperation *operation, NSError *error))failure
-//{
-//#pragma clang diagnostic push
-//#pragma clang diagnostic ignored "-Warc-retain-cycles"
-//#pragma clang diagnostic ignored "-Wgnu"
-//
-//    self.completionBlock = ^ {
-//        dispatch_async(image_request_operation_processing_queue(), ^(void) {
-//            if (self.error) {
-//                if (failure) {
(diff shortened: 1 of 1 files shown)


haihw:
Hi all,
Does `NSURLCache` work for the response that contains `Tranfers-Encoding: Chunked` and no `Content-Length` provided.

PR #2022, opened 2014-04-22
feat(AFHTTPRequestOperationManager): PUT Request w/ Constructing block

callbacknull:
A PUT request that accepts a `AFMultipartFormData` Protocol block


netwire88:
Can we reopen this discussion? I'd be very useful to have a PUT that supports construction block.

PR #2085, opened 2014-05-22
NSProgresses should be created within context of currentProgress

stigi:
There's currently an issue that prevents the progresses of upload and download tasks from being good citizens inside the NSProgress protocol.

First of all `AFURLSessionManagerTaskDelegate` creates the progress objects in its `-init` method and explicitly sets their parent to `nil`, so they'll never be added as children to `[NSProgress currentProgress]`.

The second issue is the fact that we're actually creating two progresses for upload and download. For a download task we're only updating the download progress though. The upload progress will block 50% of the pending changes in the parent progress though.

What's in this pull request:
- Tests for `AFURLSessionManager` that demonstrate the issues
- Updates to `AFURLSessionManager` to fix them

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -96,8 +96,7 @@ typedef void (^AFURLSessionTaskCompletionHandler)(NSURLResponse *response, id re
 @interface AFURLSessionManagerTaskDelegate : NSObject <NSURLSessionTaskDelegate, NSURLSessionDataDelegate, NSURLSessionDownloadDelegate>
 @property (nonatomic, weak) AFURLSessionManager *manager;
 @property (nonatomic, strong) NSMutableData *mutableData;
-@property (nonatomic, strong) NSProgress *uploadProgress;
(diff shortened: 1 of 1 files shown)


kcharwood:
I'm revisiting this now, and wondering if we actually do need to progress properties - one for upload and one for download. NSURLSessionTask splits the results out through four properties, so it makes sense to me to have two NSProgresses representing upload and download.

kcharwood:
In your case @stigi, I wonder if we should have people explicitly adding the progress to the parent progress object, since not everyone would want to track upload and download, but there are valid cases for both?

stigi:
Oh... That's an old PR :stuck_out_tongue_winking_eye:
The way `NSProgress` instances are composed is weird, I agree. I put up that PR because I wanted to support the Apple way™ of having one parent progress per thread. I didn't really distinguish between up and download at that time. I think we mostly cared about overall progress & when the overall request is expected to finish.

kcharwood:
Yep its old, but still the way things are done. My current gut is to refactor and require a user to explicitly manage and compose the parent progress to fit their needs. Do you see any objections?

PR #2252, opened 2014-08-18
Make error accessible on AFNetworkingOperationDidFinishNotification

mrackwitz:
If you observe `AFNetworkingOperationDidFinishNotification` for a `AFHTTPRequestOperation`, and the operation fails because of an serialization error (e.g. an unexpected status code), you won't be able to detect this by checking the operation's error property. This is valid as this is the intended behavior of the API.

This affects some special cases, for example AFNetworking/AFNetworkActivityLogger/issues/14.
(I have seen the related and closed issue #1968, and similars like #1876.)

But currently there are only some rather ugly ways to workaround this:
* call `[operation responseObject]`, which will initialize the object as a side-effect.
* do not rely on notifications only (e.g. use in combination `dispatch_notify` and a custom `completionGroup`), but this has in any case side-effects to the operation.

Without neccessary hurting the current API contract (which possibly depends on interpretation), there would be also the following solutions:

1. Ensure that it was tried once to deserialize the response, when `error` is called.
- Add a new method (e.g. `[operation errorEnsureResponseDeserializationWithCompletionBlock:]`) where the same as in the previous suggestion would be done, but with a block-based interface as this call could otherwise led into blocking the main thread with an expensive deserialization operation. This would ensure furthermore full backwards-compatibility.

This PR goes solution proposal number 1 and make the side-effect from responseObject *less* transparent.

### DIFF
--- a/AFNetworking/AFHTTPRequestOperation.m
+++ b/AFNetworking/AFHTTPRequestOperation.m
@@ -80,22 +80,32 @@ static dispatch_group_t http_request_operation_completion_group() {
     [self.lock unlock];
 }
 
-- (id)responseObject {
+- (void)deserializeIfNeeded {
     [self.lock lock];
-    if (!_responseObject && [self isFinished] && !self.error) {
-        NSError *error = nil;
-        self.responseObject = [self.responseSerializer responseObjectForResponse:self.response data:self.responseData error:&error];
-        if (error) {
-            self.responseSerializationError = error;
-        }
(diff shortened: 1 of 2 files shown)


kcharwood:
Given the low volume of requests for this feature, and the fact that we will soon be removing support for NSURLConnection based API's, I'm going to close this PR.

Thanks for taking the time to put it together! :beers:

PR #2372, opened 2014-10-20
[Issue #2369] Added lock/unlock to description method

mbjtrifork:


woshimaliang:
Why would it cause a crash? Just because it is customized getter method, so it broke the default "atomic" attribute? but the NSString *description = [NSString stringWithFormat:...] is dynamically created every time, still threading safe problem?

PR #2457, opened 2014-11-30
Display all of variable arguments which are specified as otherButtonTitles

totomo:
I found a minor problem at following methods which are in category extension class for UIAlertView.
 * ``` + (void)showAlertViewForTaskWithErrorOnCompletion:(NSURLSessionTask *)task
                                         delegate:(id)delegate
                                cancelButtonTitle:(NSString *)cancelButtonTitle
                                otherButtonTitles:(NSString *)otherButtonTitle, ... ```
 * ``` + (void)showAlertViewForRequestOperationWithErrorOnCompletion:(AFURLConnectionOperation *)operation
                                                     delegate:(id)delegate
                                            cancelButtonTitle:(NSString *)cancelButtonTitle
                                            otherButtonTitles:(NSString *)otherButtonTitle, ...```  

For instance, if the ```otherButtonTitles, ...``` variable argument is set to ```@"Button1", @"Button2", @"Button3", nil``` , the alertView should be displayed like this.  

UIAlertView (set to Button1, Button2, Button3 as otherButtonTitle)

However,  these methods display only first of variable arguments (``` @"Button1" ```) like this.
Specify Button1, Button2, Button3..but showed Button1 only

The reason of this, these method deliver the argument which is recieved  as "```otherButtonTitles```" to UIAlertView's function as it is.
Variable arguments cannot be delivered  to another method.
So I attempt to separate them for each argument by using ```va_start, va_arg, va_end``` API and set these separated arguments to title of ```UIAlertView```.
Hope you like this proposal :smile:

### DIFF
--- a/UIKit+AFNetworking/UIAlertView+AFNetworking.m
+++ b/UIKit+AFNetworking/UIAlertView+AFNetworking.m
@@ -62,6 +62,13 @@ static void AFGetAlertViewTitleAndMessageFromError(NSError *error, NSString * __
                                 cancelButtonTitle:(NSString *)cancelButtonTitle
                                 otherButtonTitles:(NSString *)otherButtonTitles, ... NS_REQUIRES_NIL_TERMINATION
 {
+    va_list otherTitleList;
+    va_start(otherTitleList, otherButtonTitles);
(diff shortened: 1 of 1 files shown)


mattt:
Thanks for the PR, @totomo, and apologies for taking so long to evaluate this. A new release of AFNetworking is forthcoming.

totomo:
Thanks for merging, @mattt .

Issue #2458, opened 2014-11-30
AFNetworking 2.5.0 (#2411) breaks AFNetworkActivityIndicatorManager on iOS 7


tangphillip:
My [swizzle-fix](https://github.com/AFNetworking/AFNetworking/pull/2411) for #1477 broke `AFNetworkActivityIndicatorManager` on iOS 7.
_Reported by @erikjalevik_ [here](https://github.com/AFNetworking/AFNetworking/issues/1477#issuecomment-63974854)_._

The swizzle was intended to replace KVO-watching the `NSURLSessionTask`s' `state` by producing `NSNotification`s whenever `-resume` or `-suspend` was called on any `NSURLSessionTask`.

On iOS 8, this approach works. On iOS 7, however, the public subclasses of `NSURLSessionTask` do not actually derive from a class called `NSURLSessionTask `. Instead, they derive from a class called `__NSCFLocalSessionTask`. Swizzling its `-resume` and `-suspend` methods works, and solves the issue.

But...as the name suggests, it's a private class. We can't go about calling `NSClassFromString("__NSCFLocalSessionTask")`. But could we swizzle `[taskInstance superclass]`? I'm not sure.

Code here: https://github.com/AFNetworking/AFNetworking/pull/2459

amarcadet:
Hi, any news on this?

mattt:
Fixed by #2477.

PR #2459, opened 2014-11-30
Fix AFNetworkActivityIndicatorManager for iOS 7...dubiously

tangphillip:
The AFNetworkActivityIndicatorManager is broken for iOS 7, as of #2411. (See #2458.)

This swizzle-fix for #1477 used to assume that all `NSURLSessionTask` instances had a superclass named `NSURLSessionTask`. Unfortunately, that is not true on iOS 7 devices.

We still need to somehow swizzle a superclass of all `NSURLSessionTask`s, though. On iOS 7, one of those superclasses is `__NSCFLocalSessionTask`, which is `NSURLSessionDataTask`'s superclass. This is a way to do it without explicitly mentioning or calling any private API.

However, it does end up adding and swizzling methods on `__NSCFLocalSessionTask`. Is this legal? I hope so, but ultimately, I don't know.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -272,19 +272,27 @@ static inline void af_swizzleSelector(Class class, SEL originalSelector, SEL swi
     }
 }
 
+static inline void af_addMethod(Class class, SEL selector, Method method) {
+    class_addMethod(class, @selector(af_resume),  method_getImplementation(method),  method_getTypeEncoding(method));
+}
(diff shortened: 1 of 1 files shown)


mattt:
@tangphillip Can you verify this fix against the latest master and resubmit a PR? I'd like to incorporate a dubious fix before burning the entire mechanism down as it currently exists in a more significant release.

tangphillip:
@mattt Verified and resubmitted as #2564.

PR #2479, opened 2014-12-18
Adopting modern Objective C ( Add the NS_DESIGNATED_INITIALIZER macro).

samirGuerdah:
Clarify the distinction between designated and convenience initializers. This macro will help to have warnings from the complier when the initializations pattern/restrictions are violated.


tewha:
This has some consequences; see #2555 and #2559.

PR #2480, opened 2014-12-18
Image response serializer weirdness

plu:
Hey @mattt,

first a disclaimer: This pull request is not supposed to be merged! I just kindly want to ask you, if you could have a look at it.

There are two commits. One adding a failing test. And another one fixing it. The fix is really weird and I honestly don't understand why it is fixing the issue.

Also, in our main project (eBay Kleinanzeigen), which is using AFNetworking, the malloc guard breakpoint points to https://github.com/AFNetworking/AFNetworking/blob/c3a929de8c9f6575ac5e62e0f9f14aa4ca39b93e/AFNetworking/AFURLResponseSerialization.m#L616

But I cannot reproduce that in the AFNetworking iOS Example, there the malloc guard breakpoint does not fire. But it's a different setup there anyway. It's using the category on UIImage, which is using AFHTTPRequestOperation etc.

Maybe I'm doing something completely wrong here in this NSURLSession + AFImageResponseSerializer setup? Any pointers are appreciated!

Thanks in advance,
plu

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -541,6 +541,8 @@ static UIImage * AFInflatedImageFromResponseWithDataAtScale(NSHTTPURLResponse *r
         return nil;
     }
 
+    UIImage *image = AFImageWithDataAtScale(data, scale);
+
     CGImageRef imageRef = NULL;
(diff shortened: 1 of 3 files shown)


kcharwood:
I think this is fixed with #2860!

:beers:

PR #2490, opened 2014-12-31
Added download progress block to UIImageView category

callo90:
Very useful to know the exact progress for each image download using AFNetworking UIImageView category.


mattt:
I disagree that this is generally useful. If you want any specific behavior beyond what's provided (such as in this case or a wide range of other possibilities), you can use request operations or a session manager directly.

PR #2491, opened 2015-01-01
Replace sizeWithFont with boundingRectWithSize, and remove the #pragma c...

liufan321:
Replace sizeWithFont with boundingRectWithSize, and remove the #pragma clang diagnostic ignored "-Wdeprecated-declarations"


mattt:
Thanks for the fix, @liufan321.

PR #2492, opened 2015-01-03
fix genstrings warning in UIAlertView+AFNetworking

aporat:
` find . -name \*.m | xargs genstrings -o .` generate the following warning: 

Warning: Key "Dismiss" used with multiple comments "UIAlert View Cancel Button Title" & "UIAlertView Cancel Button Title"


mattt:
Thanks, @aporat.

PR #2509, opened 2015-01-10
include responseObject in error object

CalQL8ed-K-OS:
an HTTP service may respond with a 400 series status, but contain useful error information in the body as JSON.
in this case, the default behavior of JSONResponseSerializer is to treat any non 200 series response as an error, and fill it's `error` AND `responseObject` _out_ parameters.

my issue comes with the layer introduced by the `GET:parameters:success:failure:` family of methods, which i want to use, but which assume that any errors, including those non 200s, means we shouldn't care about the responseObject.

i'm not sure that my solution here is best, but in the app i'm working on, i got my desired result by overriding behavior in my subclass as follows:
```objc
-(NSURLSessionDataTask*) dataTaskWithRequest:(NSURLRequest*) request
                           completionHandler:(void (^)(NSURLResponse*, id, NSError*)) completionHandler
{
    return [super dataTaskWithRequest: request
                    completionHandler:
            ^(NSURLResponse* response, id responseObject, NSError* error) {
                if (error != nil && responseObject != nil) {
                    error = [NSError errorWithDomain: error.domain
                                                code: error.code
                                            userInfo: ({
                        NSMutableDictionary* userInfo = [error.userInfo mutableCopy];
                        userInfo[AFNetworkingTaskDidCompleteSerializedResponseKey] = responseObject;
                        userInfo;
                    })];
                }
                
                if (completionHandler)
                    completionHandler(response, responseObject, error);
            }];
}
```

i'm pretty sure this is a legitimate use case, but i welcome any advise here

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -184,6 +184,17 @@ didCompleteWithError:(NSError *)error
 
             if (responseObject) {
                 userInfo[AFNetworkingTaskDidCompleteSerializedResponseKey] = responseObject;
+                
+                if (serializationError) {
+                    NSDictionary* userInfo =  ({
+                        NSMutableDictionary* d = serializationError.userInfo.mutableCopy;
+                        d[AFNetworkingTaskDidCompleteSerializedResponseKey] = responseObject;
+                        d;
+                    })
+                    serializationError = [NSError errorWithDomain: serializationError.domain
+                                                             code: serializationError.code
+                                                         userInfo: userInfo];
+                }
             }
 
             if (serializationError) {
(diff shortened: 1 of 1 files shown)


mattt:
The `responseObject` of a failed task is already accessible through the `AFNetworkingTaskDidCompleteSerializedResponseKey` key of the error `userInfo`.

CalQL8ed-K-OS:
hmm, that's not what I was seeing.
I'll take a look at it again, see if I'm missing something.
¿where is that response put into the error object?
I only saw it put into the Notification, not the callback block's error.

CalQL8ed-K-OS:
i'm looking at this again @mattt, and i definitely think there's still something to my pull request.
the `userInfo` variable in the `URLSession:task:didCompleteWithError:` method is only used in the __NSNotification__, and not the __NSError__ that is passed to the callbacks.
¿what do you think?

mmmilo:
@mattt +1 to this
For example, a server might return a 400 error, with some JSON that describes the error.

CalQL8ed-K-OS:
@mattt, I'd rather not make a duplicate PR, but since @mmmilo, seems to agree, ¿can we take another look at this?

PR #2511, opened 2015-01-11
MapKit category

newmanw:
The UIImageView category is awesome, I use it all the time.  I also write a lot of apps with MapKit and pull MKAnnotationVIew images from a URL.  This is exactly the same problem that the UIImageView category solves.

Added a new MKAnnotationView category that will retrieve annotation images from a URL request.  This category leverages the image cache created in the UIImageView category.


mattt:
Thanks for taking the time to create this, @newmanw. While this is probably a bit too specific for the standard library, this would be great as an AFNetworking extension. Please consider extracting this into a new project.

newmanw:
thanks @mattt.  I do disagree a little though, how is it any more specific than UIImageView.  Also takes advantage of the image cache that UIImageView is already using.

Its really a bummer that mapping applications always seem to take second fiddle.  I would argue that Maps are as prevalent as tableviews and the like (which is what the UIImageView category it used for).

mattt:
@newmanw You can't build a (non-game) iOS app without UIKit. The same is not true of MapKit. This is not to say that MapKit isn't important, it's just that including it as a core component of an already sizable dependency.

newmanw:
@mattt fair enough. I do appreciate AFNetworking a ton. And thanks for talking a look at the pill request.

PR #2521, opened 2015-01-16
Added MIME-type "image/pjpeg" to acceptable content types in AFURLRespon...

DrMephitso:
Added MIME-type "image/pjpeg" to acceptable content types in AFURLResponseSerialization


mattt:
AFNetworking only uses the official MIME types for particular content types, as specified by the IANA. The correct change would be to specify it in `acceptableContentTypes`, as per [this Stack Overflow question](http://stackoverflow.com/questions/21466609/afnetworking-2-setimagewithurl-custom-response-type-image-pjpeg)

PR #2526, opened 2015-01-22
return early if download task not created

ConfusedVorlon:
This happens sometimes in my testing on iOS 7.1 (using a real device, with a valid request)

I see that you have fixed a similar issue in the taskWithTask methods, so have fixed in the same way.
I haven't been using the upload tasks, but they could probably benefit from the same protection.

Without the early NULL, I hit an assertion failure in setDelegate:forTask:


kcharwood:
Interesting. According to the docs for `NSURLSession`, it is a nonnull return from `downloadTaskWithRequest`.

I'm going to leave this out for now, but thanks for the pr! :beers:

ConfusedVorlon:
the docs lie!

PR #2527, opened 2015-01-23
Fix validate domain name

DigiMinagawa:
Things AFSecurityPolicy validatesDomainName.
Them are not performed evaluation of hostname of the SSL certificate, I think not good.

ValidatesDomainName in PinningModeNone the "Defaults NO" So
is not performed evaluation of hostname.
I think it's almost the same as this vulnerability
http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2014-3577

```    
if (self.validatesDomainName) {
        [policies addObject:(__bridge_transfer id)SecPolicyCreateSSL(true, (__bridge CFStringRef)domain)];
    } else {
        [policies addObject:(__bridge_transfer id)SecPolicyCreateBasicX509()];
    }
```

The implementation in the "defaults YES" at initialization.
I think if you want to debug, and it is reasonable to be in "defaults NO".

What do you think?


DigiMinagawa:
awesome @reacuna, @mattt 
Thanks!
b95f00832feaadb88078a543eb6a3baa34458bc3

DigiMinagawa:
hi! @mattt

Forgive me for stealing your time...
May I have your comments on pull request?

mattt:
Thanks for your PR, @DigiMinagawa, but the existing behavior is correct. Applications should use the security policy appropriate for them.

co3k:
Don't you understand that the current implementation is danger?

> Applications should use the security policy appropriate for them.

The **SERIOUS SECURITY VULNERABILITY** has been introduced by the 5eaa869b9e2afe2719e60e72d353d9b57af5b946 commit, is contained the release of 2.1.0. This means that the users of earlier versions of 2.1.0 might append this security vulnerability to their apps if such users try upgrade to later versions of 2.1.0...

And, also, this type of library is usually expected that should be "secure default" (a.k.a. "secure by default") even more so for this topic since handling TLS connection is really security related feature.

co3k:
@mattt Please reconsider your decision. If not, I should announce about this vulnerability to more public space (e.g. full disclosure) before fix this issue.

DigiMinagawa:
hi @mattt

"SERIOUS SECURITY VULNERABILITY"
It, I am same thought as @co3k.
So I have changed the title.

PR #2528, opened 2015-01-23
make parameters accepting other types

mooosu:
parameters maybe a raw string or other things , when parameters is String , we do need twice cast in Swift  (parameters as AnyObject) as String. So, I think it's comfortable to change parameters's type to id


mattt:
I don't necessarily agree that strings as parameters is a valid use case, but I do think this change is warranted, since all of the other methods that take a `parameters` parameter were changed to accept `id`.

PR #2533, opened 2015-01-26
delete dispatch_once

CaoYuIOS:
when the app launch, the load method will be called only once, I think the disaptch_once at here could be omitted


mattt:
`dispatch_once` is a safeguard against potential side effects caused by `+load` being called directly. It is correct as-is.

PR #2557, opened 2015-02-11
add UIImage support in AFHTTPRequestSerializer.

zhfish:
add UIImage and NSImage support in AFHTTPRequestSerializer.
it can convert Image To NSData with fileName and mimeType.

in  multipartFormRequestWithMethod, UIImage will auto use fileName and mimeType
more convenient than constructingBodyWithBlock.


mattt:
Thanks for the PR, @zhfish, I don't see any particular reason why this should exist, when it is trivially easy to encode an image into data and construct a multi part request like you would for any data blob. This would increase the API surface area while only making image encoding marginally more convenient for a subset of possible image formats.

zhfish:
other data blob hasn't filename and mime,but image need.
i want send a image data blob like most data blob
Instead of using image, will feel it in addition to the dict

Issue #2555, opened 2015-02-10
2.5.0 compiles in Xcode 5, but 2.5.1 does not.


dtartaglia:
I can understand why support  for Xcode 5 would end at some point, but I was surprised that it ended in a minor version number change. Was this a mistake? (Compile error in AFURLConnectionOperation.h line 219.)

tewha:
Possibly relevant link: http://stackoverflow.com/questions/24119896/ns-designated-initializer-expected-colon

dtartaglia:
Yes, this is the issue. Yes, I can work around it. And yes, you are right.

mattt:
Fixed by #2559

PR #2559, opened 2015-02-12
Add NS_DESIGNATED_INITIALIZER for earlier versions of Xcode.

tewha:
`NS_DESIGNATED_INITIALIZER` is not included in Xcode 5. Since AFNetworking doesn't have a common #import for all headers, I just inserted a definition for it into each file that uses it. A common header would be better, of course!

I'm submitting this primarily so that people using Xcode 5 can check it. I'm not anymore. You might want to add a common header instead, or even drop support for Xcode 5. Whatever. :) Something _like_ this is necessary if Xcode 5 compatibility is to be maintained.

See: http://stackoverflow.com/questions/24119896/ns-designated-initializer-expected-colon and issue #2555.

### DIFF
--- a/AFNetworking/AFHTTPRequestOperationManager.h
+++ b/AFNetworking/AFHTTPRequestOperationManager.h
@@ -36,6 +36,14 @@
 #import "AFSecurityPolicy.h"
 #import "AFNetworkReachabilityManager.h"
 
+#ifndef NS_DESIGNATED_INITIALIZER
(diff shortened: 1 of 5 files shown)


mattt:
This is the correct fix. Thanks for the PR, @tewha. Apologies to anyone who was bit by the Xcode 5 compatibility issue. I hadn't fully considered the implications of annotating with `NS_DESIGNATED_INITIALIZER` when that PR was merged.

PR #2564, opened 2015-02-15
Fix AFNetworkActivityIndicatorManager for iOS 7...dubiously

tangphillip:
This is a re-opening of #2459, as requested by @mattt [here](https://github.com/AFNetworking/AFNetworking/pull/2459#issuecomment-74393321).

----

The AFNetworkActivityIndicatorManager is broken for iOS 7, as of #2411. (See #2458.)

This swizzle-fix for #1477 used to assume that all `NSURLSessionTask` instances had a superclass named `NSURLSessionTask`. Unfortunately, that is not true on iOS 7 devices.

We still need to somehow swizzle a superclass of all `NSURLSessionTask`s, though. On iOS 7, one of those superclasses is `__NSCFLocalSessionTask`, which is `NSURLSessionDataTask`'s superclass. This is a way to do it without explicitly mentioning or calling any private API.

However, it does end up adding and swizzling methods on `__NSCFLocalSessionTask`. Is this legal? I hope so, but ultimately, I don't know.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -272,19 +272,27 @@ static inline void af_swizzleSelector(Class class, SEL originalSelector, SEL swi
     }
 }
 
+static inline void af_addMethod(Class class, SEL selector, Method method) {
+    class_addMethod(class, @selector(af_resume),  method_getImplementation(method),  method_getTypeEncoding(method));
+}
+
 static NSString * const AFNSURLSessionTaskDidResumeNotification  = @"com.alamofire.networking.nsurlsessiontask.resume";
 static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofire.networking.nsurlsessiontask.suspend";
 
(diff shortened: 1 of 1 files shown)


mattt:
Thanks, @tangphillip!

AFNetworking/AFURLSessionManager.m
```diff
@@ -272,19 +272,27 @@ static inline void af_swizzleSelector(Class class, SEL originalSelector, SEL swi
     }
 }
 
+static inline void af_addMethod(Class class, SEL selector, Method method) {
+    class_addMethod(class, @selector(af_resume),  method_getImplementation(method),  method_getTypeEncoding(method));
```
dmcrodrigues: I believe there's a typo in here, instead of `@selector(af_resume)` I believe you should be using the `selector` parameter otherwise you will not add the specified selector, in this case `af_suspend`, to the class.

tangphillip: PR'ed as https://github.com/AFNetworking/AFNetworking/pull/2577 .

PR #2571, opened 2015-02-25
fixes bad log ordering with multiple requests

pabloroca:
When issuing multiple network calls the notification sometimes comes after the next request, so AFNetworkActivityLogger logs in wrong order

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -164,14 +164,14 @@ didCompleteWithError:(NSError *)error
     if (error) {
         userInfo[AFNetworkingTaskDidCompleteErrorKey] = error;
 
+        dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


mattt:
No, the current behavior is correct. Notifications should be sent out after the completion block begins, which means dispatching within the completion group.

pabloroca:
The current behavior leads to fail if we have many requests. Let's say we do this:

request A
request B (using some data retrieved from request A)

The process works as expected, but not the log.

Expected

log request A
log data received from request A
log request B

Observed:

log request A
log request B
log data received from request A
...

I created another pull request and moved the postNotification right just before the completionHandler and inside the dispatch_group_async. Tested with the same environment and works as expected

https://github.com/AFNetworking/AFNetworking/pull/2574

mattt:
@pabloroca What you described is exactly the desired and expected behavior. Notifications are sent off after the completion block fires. If you're sending a new request in a completion handler, that's just what will happen.

PR #2574, opened 2015-02-28
Fixes bad log order. completionHandler can make a new request before log results from previous request

pabloroca:


pabloroca:
The current behavior leads to fail if we have many requests. Let's say we do this:

request A
request B (using some data retrieved from request A)

The process works as expected, but not the log.

Expected

log request A
log data received from request A
log request B

Observed:

log request A
log request B
log data received from request A
...

This pull request moves the postNotification right just before the completionHandler and inside the dispatch_group_async. Tested with the same environment and works as expected

mattt:
See #2571.

PR #2581, opened 2015-03-05
Move details of XML and JSON parsing into separate virtual methods

rvangraan:
It is not possible today to inherit from AFJSONResponseSerializer and to
easily replace the JSON parser with something else (i.e. SwiftyJSON). This
PR moves the details of XML and JSON parsing (i.e. the exact class) into
separate methods that can be overloaded so that the actual JSON parsing
can be done with another library.

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.h
+++ b/AFNetworking/AFURLResponseSerialization.h
@@ -132,6 +132,16 @@
  */
 + (instancetype)serializerWithReadingOptions:(NSJSONReadingOptions)readingOptions;
 
+/**
+ Virtual method that provides the opportunity to replace the default JSON Parser
+ 
+ @param data The binary data to parse as JSON.
+ @param error An error object
+ */
+
+-(id)generateJSONResponseObject:(NSData *)data error:(NSError *__autoreleasing *)serializationError;
+
+
 @end
 
 #pragma mark -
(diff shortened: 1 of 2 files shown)


mattt:
Thanks for your PR, @rvangraan.

> It is not possible today to inherit from AFJSONResponseSerializer and to
easily replace the JSON parser with something else (i.e. SwiftyJSON).

`AFJSONResponseSerializer`, et al, are not designed to be subclassed. They are merely provided as a convenience for the majority of folks using the built-in Foundation classes. If you want to use another mechanism to process a server response, you would create a new class conforming to `AFURLResponseSerialization`. A good example of this is [AFOnoResponseSerializer](https://github.com/AFNetworking/AFOnoResponseSerializer), which processes XML using [Ono](https://github.com/mattt/Ono) rather than `NSXMLParser`.

rvangraan:
Hi,
Thanks. Making a new serializer was my first thought. 

But after looking at AFJSONResponseSerializer, I noticed a lot of logic such as this:
https://github.com/AFNetworking/AFNetworking/blob/4d635e38e20cf0a9fc178729aed76c6a524050b8/AFNetworking/AFURLResponseSerialization.m#L234-L260 AND
https://github.com/AFNetworking/AFNetworking/blob/4d635e38e20cf0a9fc178729aed76c6a524050b8/AFNetworking/AFURLResponseSerialization.m#L265-L285

that will be exactly the same in the other class. 

And most of that can't easily be reimplemented in Swift, for example:
https://github.com/AFNetworking/AFNetworking/blob/4d635e38e20cf0a9fc178729aed76c6a524050b8/AFNetworking/AFURLResponseSerialization.m#L234-L260

mattt:
@rvangraan That code is entirely specific to using `NSJSONSerialization`. You shouldn't be reimplementing that for another JSON parser.

Issue #2582, opened 2015-03-05
AFHTTPRequestSerializer returns request with incorrect timeoutInterval


micahhainline:
AFHTTPRequestSerializer.requestWithMethod:URLString:parameters:error: returns an NSMutableURLRequest with the default timeoutInterval set, instead of respecting the timeoutInterval that was set on the serializer.

danielgomezrico:
+1 Same for me

mattt:
Duplicate of #2563.

danielgomezrico:
is it a duplicate? 

I'm not subclassing AFHTTPRequestOperationManager and it happens too.

mattt:
Yes, it's the same underlying cause.

danielgomezrico:
Ok, I didn't said anything ;)

PR #2583, opened 2015-03-05
Added a fix for issue #2582

micahhainline:


mattt:
Thanks, @micahhainline, but this is not the correct resolution for this bug. It all has to do with the lines after the diff dealing with observed key paths. This will be resolved in the next release.

PR #2589, opened 2015-03-11
Guard for unsupported features of iOS 8 extensions

phylor:
See Issue #2119.

It seems that a couple of references to unsupported APIs in extensions were missed. I'd really appreciate a fix because I can't properly build my extension at the moment.

Credit goes to @jpmhouston.

### DIFF
--- a/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
+++ b/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
@@ -114,7 +114,9 @@ static NSURLRequest * AFNetworkRequestFromNotification(NSNotification *notificat
 }
 
 - (void)updateNetworkActivityIndicatorVisibility {
+#if !defined(AF_APP_EXTENSIONS)
     [[UIApplication sharedApplication] setNetworkActivityIndicatorVisible:[self isNetworkActivityIndicatorVisible]];
+#endif
 }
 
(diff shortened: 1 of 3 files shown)


mattt:
Thanks for the PR, @phylor. I missed that diff from @jpmhouston.

kevinrenskers:
When will this be released to cocoapods?

sddhhanover:
+1

maciekish:
This doesn't appear to work for me. I'm on `master` and i can see checks for `AF_APP_EXTENSION` in your code @mattt yet the build still fails with this message. I added a pch-file to my App Extension:

    #define AF_APP_EXTENSION
    #import <HittaServiceLib/HITHittaServiceLib.h>

maciekish:
To fix this issue i had to put `AF_APP_EXTENSIONS=1`as a **preprocessor macro** for the **Pods-hitta.se WatchKit Extension-AFNetworking** target. Adding `#define AF_APP_EXTENSIONS` to the prefix header didn't work, neither did adding it as a preprocessor macro to the watch extension target work.

@mattt Have you considered using this instead? I replaced your `#if !defined(AF_APP_EXTENSIONS)` with the below code instead and it appears to be working correctly.

    #if !(defined(__has_feature) && __has_feature(attribute_availability_app_extension))
        Code that can't compile for WatchKit.
    #endif

avocade:
+1

Issue #2560, opened 2015-02-13
Can't validate domain name without using public key nor pinned certificates?


jcayzac:
In AFSecurityPolicy:

```objc
    NSMutableArray *policies = [NSMutableArray array];
    if (self.validatesDomainName) {
        [policies addObject:(__bridge_transfer id)SecPolicyCreateSSL(true, (__bridge CFStringRef)domain)];
    }
    [...]
    if (self.SSLPinningMode != AFSSLPinningModeNone && !AFServerTrustIsValid(serverTrust) && !self.allowInvalidCertificates) {
        return NO;
    }
    [...]
    switch (self.SSLPinningMode) {
        case AFSSLPinningModeNone:
            return YES; <<< ?
```

…why do you always return `YES` even if `validatesDomainName` is set? IMHO this last line should read 
`return AFServerTrustIsValid(serverTrust)`, with `serverTrust` only containing the domain policy set above.

infimache:
I have similar issue with current implementation of <code>-evaluateServerTrust:forDomain:</code>

When <code>SSLPinningMode == AFSSLPinningModeNone</code>, no validation of server trust is performed and <code>-evaluateServerTrust:forDomain:</code> always returns <code>YES</code>.

I believe issue was introduced by commit <link> https://github.com/AFNetworking/AFNetworking/commit/104ce04105098466ea0ea4e337af554d7b9df195</link>.
It states that change was performed due to failing test, but for me tests succeed with or without the change.

Proposed solution is to revert mentioned commit.

jcayzac:
I agree 104ce04 looks like a bad solution for the test it was supposed to fix.

jcayzac:
Note that the security implication for this is that by default AFNetworking happily connects to https servers with invalid certificates. The company I work for has several dozen apps that just had to be updated to work around this ーit's an e-commerce company, so a few people there were quite stressed as you can guess :dancers:

mattt:
Fixed by 90bd342774b312c86352703e04969c561643c99e. See #2611 for discussion.

peterpaulis:
to go around this add

<code>
        // security policy
        {
            AFSecurityPolicy* policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
            [policy setValidatesDomainName:NO];
            [policy setAllowInvalidCertificates:!SERVICE_FORCE_HTTPS];
            self.requestOperationManager.securityPolicy = policy;
        }
</code>

Issue #2549, opened 2015-02-05
Why isn't AFSecurityPolicy validatesDomainName enabled default?  


jmkk:
AFSecurityPolicy's "validatesDomainName" is set to NO by default, and only when using policy created from +policyWithPinningMode: does this flag get enabled.

What is the reason for this flag to default to NO? It appears that all SSL/TLS traffic should evaluate DN by default, and only for cases explicitly specified by the dev should it be changed to NO. It should be treated similar to the "validatesCertificateChain" flag.

Note that there are places where it's not possible to enforce enabling "validatesDomainName", like request operations created inside of UIImageView+AFNetworking and UIButton+AFNetworking. These requests will never follow the security policy set on the HTTP client/request manager.

mattt:
The default security policy is `AFSSLPinningModeNone`, and having `validatesDomainName` set to `NO` is consistent with this. If you require security evaluation for remotely loading images, do not use the built-in UIKit categories.

jmkk:
@mattt thanks for the response. However, AFSSLPinningModeNone for SSLPinningMode is a setting for certificate pinning. I'm not talking about certificate pinning, but regular SSL/TLS trust evaluation. 

Not validating the name in the SSL certificate against the hostname during SSL traffic is considered to be an extremely bad practice, exposing all AFNetworking clients to a potential Main in the Middle attack.

As is, AFNetworking is altering the security policy setting, which directly violates Apple's API guidelines:

> Do not disable hostname checking by creating a non-TLS policy or passing in a NULL pointer for the  hostname. If you are intentionally connecting to a host using a hostname other than one of the names listed on its certificate, you should allow the operation only if the certificate you receive from that host is valid for some other domain that you control.

(https://developer.apple.com/library/mac/documentation/NetworkingInternet/Conceptual/NetworkingTopics/Articles/OverridingSSLChainValidationCorrectly.html)

Also note that due to TLS Session Cache that takes place in SecureTransport layer (https://developer.apple.com/library/ios/qa/qa1727/_index.html), if the AFNetworking client hosts images on the same server (IP+name+port) as their other content that needs to be protected, if the first request is to fetch an image (without validating the name), the remaining calls could suffer from MITM attack for the next 10 minutes as the same (not fully protected) SSL session would be reused by iOS. 

Given the above, I strongly encourage AFNetworking to follow the established security standards and default to the more secure settings, and allow the users to opt-out as needed for testing/debugging.

PR #2592, opened 2015-03-12
Added test and implementation for self signed cert against AFSSLPinningMo..

### LINKED ISSUE #2590: Security hole / configuration problem in AFSecurityPolicy.m
[Introduced in the line below, between 2.5.0 and 2.5.1; tested on iOS 8.1.3]

When establishing a security policy for a connection you have two basic options: "**pinned**" or "**not pinned**" (default). Selecting pinned mode will then force **validatesDomainName** to **NO**.

When it comes to making a decision on whether to trust a domain there is one path that goes as follows:

    if (self.SSLPinningMode != AFSSLPinningModeNone && !AFServerTrustIsValid(serverTrust) && !self.allowInvalidCertificates) {
        return NO;
    }

  ...

    switch (self.SSLPinningMode) {
        case AFSSLPinningModeNone:
            return YES;

So if you disable Pinning mode it will **ALWAYS accept the certificate**. The only way to establish that a certificate is invalid is to set up SSLPinningMode[something, not 'None'].

So the default policy is validatesDomainName = YES, SSLPinningModeNone. This set up results in **ALWAYS accepting certificates**.

I have verified that a malicious proxy server can sniff all the contents of HTTPS communication in this case.

Please make sure this confusing state of affairs doesn't make it into production applications -- I would love to write a test / fix to help with this but I'm simply not aux fait enough with the libraries involved.

duttski:
deNone

This change updates against a change made in 2.5.1 (from 2.5.0) which significantly changed the default behaviour to act much less securely than the previous implementation. Without this change it is trivial to perform MITM attacks on users running apps utilising this library (e.g. any coffee shop in the world could read everybody’s private messages).

This fixes #2590 and also seems to answer very sensible concerns raised in:
#2560
#2549
#2527

### DIFF
--- a/Tests/Tests/AFSecurityPolicyTests.m
+++ b/Tests/Tests/AFSecurityPolicyTests.m
@@ -514,10 +514,17 @@ static SecTrustRef AFUTTrustWithCertificate(SecCertificateRef certificate) {
     XCTAssert(policy.SSLPinningMode==AFSSLPinningModeNone, @"Default policy is not set to AFSSLPinningModePublicKey.");
 }
 
-- (void)testDefaultPolicyIsSetToNotAllowInvalidSSLCertificates {
-    AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
+- (void)testDefaultPolicySetToNoneValidatesDomainFailsWithBadCertificate {
+    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
+    SecTrustRef trust = AFUTTrustWithCertificate(AFUTSelfSignedCertificateWithoutDomain());
+    XCTAssert([policy evaluateServerTrust:trust forDomain:nil] == NO, @"Unmatched certificate and pinning mode None should fail");
+    CFRelease(trust);
+}
 
-    XCTAssert(policy.allowInvalidCertificates == NO, @"Default policy should not allow invalid ssl certificates");
+- (void)testDefaultPolicyIsSetToAFSSLPinningModeNone {
+    AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
+    
+    XCTAssert(policy.SSLPinningMode==AFSSLPinningModeNone, @"Default policy is not set to AFSSLPinningModeNone.");
 }
 
 - (void)testPolicyWithPinningModeIsSetToNotAllowInvalidSSLCertificates {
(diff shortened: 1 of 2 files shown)


Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -514,10 +514,17 @@ - (void)testDefaultPolicyIsSetToAFSSLPinningModePublicKey {
     XCTAssert(policy.SSLPinningMode==AFSSLPinningModeNone, @"Default policy is not set to AFSSLPinningModePublicKey.");
 }
 
-- (void)testDefaultPolicyIsSetToNotAllowInvalidSSLCertificates {
-    AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
+- (void)testDefaultPolicySetToNoneValidatesDomainFailsWithBadCertificate {
+    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeNone];
```
kcharwood: This test is misnamed. You aren't testing the default policy, you are testing a policy with pinning mode set to `AFSSLPinningModeNone`.

kcharwood: Since there is another test to verify the `[AFSecurityPolicy defaultPolicy]` is set to `AFSSLPinningModeNone`, you can probably just safely use `[AFSecurityPolicy defaultPolicy]` and rename the test to `testDefaultPolicyFailsToEvaluateSelfSignedCertificate`

duttski: I agree with the name change, and would really like someone to take a look at the test too, but it seems to be being ignored on the whole. I will have a go at improving this today. 

What I was after was a test that says:
    
   "If the developer doesn't specify security behaviour we fall back onto a sensible security policy"

However, since the ```AFSSLPinningModeNone``` represents an absence of ```[pinning modes]``` it doesn't seem appropriate to simply have two tests that show ```default== AFSSLPinningModeNone``` && ```AFSSLPinningModeNone==[some level of security]```, since it's entirely reasonable that ```[some level of security]``` could change over time, whereas we don't want the default security behaviour to be so variable.

Given the number of apps using AFNetworking I think this is a major concern -- how many of my password are being easily sniffed by coffee shops!?!

duttski: @kcharwood  I created a separate pull request: https://github.com/AFNetworking/AFNetworking/pull/2611

PR #2598, opened 2015-03-16
Fix AFImageWithDataAtScale destroying animation images

petrocket:
AFImageWithDataAtScale destroys animated UIImages because initWithCGImage only uses the first frame.  This patch fixes the issue, though perhaps there is some way to scale each individual frame which would could be a huge CPU/mem drag.


mattt:
Merged with fe0ddd06d8ad7d957c0e60cc9a3fff65d8c260b9. Thanks, @petrocket.

Issue #2605, opened 2015-03-20
UIWebView+AFNetworking.m crash


dake:
line 140
[strongSelf.delegate webViewDidFinishLoad:strongSelf]; 
should change to 
 if ([strongSelf.delegate respondsToSelector:@selector(webViewDidFinishLoad:)]) {
        [strongSelf.delegate webViewDidFinishLoad:strongSelf];
  }

line 150
[self.delegate webViewDidStartLoad:self];
should change to 
 if ([self.delegate respondsToSelector:@selector(webViewDidStartLoad:)]) {
        [self.delegate webViewDidStartLoad:self];
 }

mattt:
Please submit a PR with these changes.

PR #2610, opened 2015-03-25
fix UIWebView+AFNetworking crash

dake:
fix issue #2605


mattt:
Thanks, @dake! I really appreciate it.

PR #2611, opened 2015-03-25
Added test and implementation for self signed cert against default secur...

### LINKED ISSUE #2590: Security hole / configuration problem in AFSecurityPolicy.m

duttski:
...ity

This change updates against a change made in 2.5.1 (from 2.5.0) which significantly changed the default behaviour to act much less securely than the previous implementation. Without this change it is trivial to perform MITM attacks on users running apps utilising this library (e.g. any coffee shop in the world could read everybody’s private messages).

This fixes #2590 and also seems to answer very sensible concerns raised in:
#2560
#2549
#2527

### DIFF
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -254,14 +254,21 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
 
     SecTrustSetPolicies(serverTrust, (__bridge CFArrayRef)policies);
 
-    if (self.SSLPinningMode != AFSSLPinningModeNone && !AFServerTrustIsValid(serverTrust) && !self.allowInvalidCertificates) {
+    if(self.SSLPinningMode == AFSSLPinningModeNone) {
+        if(self.allowInvalidCertificates || AFServerTrustIsValid(serverTrust)){
+            return YES;
+        } else {
+            return NO;
+        }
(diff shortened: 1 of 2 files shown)


mattt:
Thanks for this, @duttski. A new release with this patch is forthcoming.

duttski:
@mattt no problem. I hope [someone] has had chance to check the tests and implementation. I've not been looking at AFNetworking long so it's probably worth a second opinion (especially given how deep it is in the library).
Regards,

PR #2613, opened 2015-03-26
1

AlexYang1949:
1


Legoless:
Nice description, care to elaborate?

mattt:
This appears to be adding documentation in Chinese. Although I appreciate the effort, adding an additional localization to documentation is a serious undertaking. Right now, I don't think we're prepared to fully commit to ensuring that docs remain accurate and self-consistent across translations.

PR #2620, opened 2015-03-27
Make spacing for if statements consistent

streeter:


mattt:
:+1: 

Thanks, @streeter.

PR #2621, opened 2015-03-28
Fix dyld startup crash on iOS 6 introduced in 2.5.2

btgerst:
The use of `NSURLSessionTask` concrete subclasses (i.e. `NSURLSessionDataTask`) [at load time](https://github.com/AFNetworking/AFNetworking/blob/a13a244526e4110739872904c860edd645accdd1/AFNetworking/AFURLSessionManager.m#L287) was causing `dyld` crashes on iOS 6 devices.  I had to do some experimenting in the `AppDelegate` of **iOS Test Application** to figure out what exactly was causing the `dyld` crash.  This is what led to me making the following changes:

### Modifications

- Swizzling the `resume` and `suspend` implementations happens in a category on `NSURLSessionTask` instead of `NSURLSessionDataTask`.  `NSURLSessionTask` is weakly linked via availability macros, which makes it safe to hang the swizzling on.
- Removed `dispatch_once` wrapper as `+load` is only invoked once.
- Added a (logic) test for the `resume` and `suspend` notifications

### Additions

- **`iOS Test Application`**: iOS application test target for running tests on physical devices.
- **`AFURLSessionManager`**: 
  - `+(BOOL)isAvailable`: checks that `NSURLSession` is available as a convenience for backwards-compatible and SDK-dependent development
- **`AFTestCase`**:
  - `+(BOOL)requiresSessionAPIAvailability`: whether the test class should only be run if `NSURLSession` is available
  - `-(void)setUpSessionTest`: setup method which is only invoked when `NSURLSession` is available
  - `-(void)tearDownSessionTest`: same as above, but for `tearDown`.

Tested myself by running the test application and application tests on my iOS 6 device.

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.m
+++ b/AFNetworking/AFHTTPSessionManager.m
@@ -22,7 +22,7 @@
 
 #import "AFHTTPSessionManager.h"
 
-#if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
+#if (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
 
 #import "AFURLRequestSerialization.h"
 #import "AFURLResponseSerialization.h"
--- a/AFNetworking/AFHTTPSessionManager.m
+++ b/AFNetworking/AFHTTPSessionManager.m
@@ -22,7 +22,7 @@
 
 #import "AFHTTPSessionManager.h"
 
-#if (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
+#if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
 
 #import "AFURLRequestSerialization.h"
 #import "AFURLResponseSerialization.h"
--- a/AFNetworking/AFURLSessionManager.h
+++ b/AFNetworking/AFURLSessionManager.h
@@ -188,6 +188,13 @@
  */
 @property (nonatomic, assign) BOOL attemptsToRecreateUploadTasksForBackgroundSessions;
 
+///-----------------------
+/// @name API Availability
+///-----------------------
+
+/// @return @c NO if the API is unavailable due to required classes not being present (i.e. on iOS 6).
++ (BOOL)isAvailable;
+
 ///---------------------
 /// @name Initialization
 ///---------------------
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -23,7 +23,7 @@
 #import "AFURLSessionManager.h"
 #import <objc/runtime.h>
 
-#if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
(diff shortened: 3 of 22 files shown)


dr0ptp4kt:
:+1:

AFNetworking/AFHTTPSessionManager.m
```diff
@@ -22,7 +22,7 @@
 
 #import "AFHTTPSessionManager.h"
 
-#if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
+#if (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
```
kylef: I don't think this change is correct, since this API should still be accessible when building an application supporting both iOS 6 and iOS 7. A developer can conditionally use this when the user's device is running iOS 7 or greater.

I.e, this API should still be available when the minimum required version is iOS 6 because I can conditionally use it.

clemensg:
Same issue here when running an application with AFNetworking 2.5.2 on OS X 10.8 ...

AFNetworking/AFHTTPSessionManager.m
btgerst: @kylef I see your point, but I had thought the intent of the macros was to disable compilation when the deployment target was iOS 6 (which is indicated by `__IPHONE_OS_VERISON_MIN_REQUIRED`, not `MAX_ALLOWED`).

Would the preferred approach be to wrap [the offending code](https://github.com/AFNetworking/AFNetworking/blob/a13a244526e4110739872904c860edd645accdd1/AFNetworking/AFURLSessionManager.m#L287) in conditionals which check for class availability at runtime?

I guess it depends on the maintainers' views on safety vs. flexibility.  I don't know of any reliable ways to trigger class availability warnings at compile time, though [some seem to have had some success](http://stackoverflow.com/a/8993138/600467).

kylef: I think that would be a far better option, this was introduced in #2411. I'm not entirely sure if it would work, since this is a category on a class that won't be available on iOS 6.

If it doesn't work, we could move the method swizzling to a function such as the following:

```
__attribute__((constructor)) static void AFNetworkingSwizzleNSURLSessionTask(void) {
    if (NSClassFromString(@"NSURLSessionTask")) {
        // continue exchanging NSURLSessionTask's methods
    }
}
```

Note, I think it would be a good idea to make the unit-tests for AFNetworking ran by travis also run the tests on iOS 6 to ensure this kind of thing doesn't happen again.

btgerst: > Note, I think it would be a good idea to make the unit-tests for AFNetworking ran by travis also run the tests on iOS 6 to ensure this kind of thing doesn't happen again.

:+1:, but it seems [Travis doesn't support iOS 6](http://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions)?

clemensg: Running the tests on 10.8 would also ensure that this does not happen again.

btgerst:
@kylef this new approach (see above description) should be safe while still allowing people to conditionally use the session-based APIs.  Let me know if you have any questions while I look into the test failure (one of which isn't my fault, AFAIK).

AFNetworking/AFURLSessionManager.h
```diff
@@ -188,6 +188,13 @@
  */
 @property (nonatomic, assign) BOOL attemptsToRecreateUploadTasksForBackgroundSessions;
 
+///-----------------------
+/// @name API Availability
+///-----------------------
+
+/// @return @c NO if the API is unavailable due to required classes not being present (i.e. on iOS 6).
++ (BOOL)isAvailable;
```
kylef: IMO, this should be a property:

```objective-c
@property (nonatomic, assign, readonly, getter=isAvailable) available;
```

btgerst: I disagree since you want to do the check *before* creating an instance of the class, and you don't need an instance to do the check.

btgerst:
I think the Travis agent bought the farm.  The tests pass for me on my machine.  @kylef can you confirm locally and/or retrigger a build?

AFNetworking/AFURLSessionManager.h
kylef: Sorry, mis-read it was a class method, the `+` looked part of the diff in my tired eyes.

Make's sense :+1:

btgerst:
@kylef anything else that needs to be done to get this merged?

kylef:
@btgerst I'm not actually an owner of this project, merging would be down to the owners.

The pull request looks good to me :+1:.

mattt:
I apologize for the oversight, @btgerst, and appreciate your PR. However, the correct fix should be much more straightforward—only conditionally swizzle the class if it is available, using the `NS_CLASS_AVAILABLE` macro by checking `if ([NSURLSessionDataTask class]) {...}`. c9f8d1e6211d12aff7e1d3237db9ecaf81a18a69 makes this change.

Can you confirm this fix?

btgerst:
@mattt not sure if you saw the [comments on the patch](https://github.com/AFNetworking/AFNetworking/commit/c9f8d1e6211d12aff7e1d3237db9ecaf81a18a69#commitcomment-10567391), but it seems like the issue is still present on iOS 6 and 10.8.

shingo-minamiyama:
I have same problem on iOS 6.1.4.
This runtime error occurred.

EXC_BREAKPOINT (code=EXC_ARM_BREAKPOINT, subcode=0xe7ffdefe)

Issue #2618, opened 2015-03-27
Reachability gives different results in 2.5.2


KristopherGBaker:
I just updated to 2.5.2 through pod update.  I noticed AFNetworking's reachability was giving me unexpected results, so I investigated further by downgrading to 2.5.1, where I saw what I normally expect.

To reproduce this issue, create a new Single View Application iOS project (Objective-C) in Xcode (I'm using Xcode 6.2 if that matters).  Setup the project to use AFNetworking 2.5.2.

Modify AppDelegate.m so it looks like below, run the project.  You should see two log statements, one saying "Network Reachable", and the other saying "Network Not Reachable".  The order of these two statements isn't always the same.  

Now setup the project to use AFNetworking 2.5.1 and run the project again.  You should see a single log statement saying "Network Reachable", which is what I would normally expect (unless I have actual network or server issues of course).

```objectivec
#import "AppDelegate.h"
#import "AFNetworkReachabilityManager.h"

@interface AppDelegate ()

@property (strong, nonatomic) AFNetworkReachabilityManager *networkReachabilityManager;

@end

@implementation AppDelegate


- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    
    self.networkReachabilityManager = [AFNetworkReachabilityManager managerForDomain:@"google.com"];
    
    [self.networkReachabilityManager setReachabilityStatusChangeBlock:^(AFNetworkReachabilityStatus status) {
        
        switch (status) {
            case AFNetworkReachabilityStatusReachableViaWiFi:
            case AFNetworkReachabilityStatusReachableViaWWAN:
                NSLog(@"Network Reachable");
                break;
                
            case AFNetworkReachabilityStatusNotReachable:
                NSLog(@"Network Not Reachable");
                break;
                
            case AFNetworkReachabilityStatusUnknown:
                NSLog(@"Reachability Status Unknown");
                
            default:
                break;
        }
    }];
    
    [self.networkReachabilityManager startMonitoring];
    
    return YES;
}

- (void)applicationWillResignActive:(UIApplication *)application {
}

- (void)applicationDidEnterBackground:(UIApplication *)application {
}

- (void)applicationWillEnterForeground:(UIApplication *)application {
}

- (void)applicationDidBecomeActive:(UIApplication *)application {
}

- (void)applicationWillTerminate:(UIApplication *)application {
}

@end
```

KristopherGBaker:
[...]

PR #2625, opened 2015-04-01
Bugfix: Incorrect ordering of reachability states (#2618)

KingOfBrian:
This fixes a reachability regression in 2.5.2 #2618

The background queue usage in the initial reachability status check was causing the reachability states to be reported out of order.  The hop onto the background queue caused the first value checked (0) to be reported after the runloop checked value (2).

I'm not sure how ok you are with things changing, so I made 3 commits for each change I think should be made. Pick the ones you want, I tested that 1 resolves the issue, 2 should also guard against the issue, and 3 is a small cleanup.

1) Remove the background queue for the initial check. There's no valid reason to use a background thread here, since all reporting and probing of reachability occurs on the main thread.
2) Check `networkReachabilityStatus` in the initial check and only trigger if it's `AFNetworkReachabilityStatusUnknown`.
3) Remove the `dispatch_async` call in `AFNetworkReachabilityCallback`. The callback and the notification were not split up across operations anywhere else, and it's purpose was not clear. This doesn't fix the bug, or really matter, but it confused me.

Thanks!

Brian


kcharwood:
This was fixed by a revert in 2.5.3. Please let me know if you continue to see issues.

Thanks for the PR!

PR #2630, opened 2015-04-02
fix for broken test

chillpop:
It appears to be a copy paste error from the test above (
line 533 of `testPolicyWithPinningModeIsSetToNotAllowInvalidSSLCertificates`).


chillpop:
This test is causing Travis to fail for PR #2625. I'm really interested in fixing the Reachability issue #2618 and thinking maybe this failing test is preventing that PR from being approved.

PR #2632, opened 2015-04-02
Fix for broken test

chillpop:
It appears to be a copy paste error from the test above (
line 533 of `testPolicyWithPinningModeIsSetToNotAllowInvalidSSLCertificates`).


mattt:
Thanks, @chillpop.

PR #2637, opened 2015-04-03
Adding default security mode test to cover main settings

### LINKED ISSUE #2631: Unable to satisfy requirements
I updated the pods gem, removed my Pods folder, then attempted pod install

```
platform :ios, '8.0'
pod 'RestKit', '~> 0.20.0pre'
pod 'SVProgressHUD'
pod 'AFNetworking', '~> 2.5'
pod 'SBJson'
pod 'JSONKit'
```

> pod install

```
[!] Unable to satisfy the following requirements:

- `AFNetworking (~> 2.5)` required by `Podfile`
- `AFNetworking (~> 1.3.0)` required by `RestKit/Network (0.20.3)`

```

### LINKED ISSUE #2633: Adding missing tests for default security mode
I am trying to contribute some tests that will ensure that the security mode established by default is to a certain standard and that there are no regressions in the future. It is also in response to the fact that in general the tests are sometimes misnamed or have mis-matching messages, which I guess could lead to issues born out of confusion.

This is in reference to two commits:
  - 8cfb272726fea72c3632e2a78638aecfeece8839 -- don't default to letting everything through
  - 3e631b203dd95bb82dfbcc2c47a2d84b59d1eeb4 -- don't default to letting invalid domains through

Which in turn have the following impact:
  - 2.5.1 fails to check anything by default
  - 2.5.2 fails to match the domain to the provided certificate by default
  - **(current master) matches both but has no tests to verify this**

As things stand there is a test for `testDefaultPolicyFailsToEvaluateServerTrustFromSelfSignedCertificate` -- I think this test should check that the system is defaulting to the 'normal' security standard, i.e. that which is normally implemented in the browser. So in total it should:
  - check that a self-signed certificate fails
  - check that a certificate with a valid route to a trusted CA root but for a different domain fails
  - check that a certificate with a valid route to a trusted CA root for a matching domain passes

At the moment:
  - (there is one failing test around the pinning mode `testPolicyWithPinningModeIsSetToValidatesDomainName` -- I'm not touching this in favour of leaving to someone who understands that area better **Edit : I think this has now been resolved**)
  - there is a test `testDefaultPolicyFailsToEvaluateServerTrustFromSelfSignedCertificate` that will pass, with the `validatesDomainName` set to `YES` or `NO`
  - I can make the tests pass by adding a line of code that ensures the `validatesDomainName` flag is NO with `AFSSLPinningModeNone` (admittedly this would be a stupid line of code to add, but nevertheless the tests don't protect against this mistake, therefore (IMO) exposing a potential regression).

In my opinion there is a(n/ strong) argument for containing (or otherwise grouping) these tests, thus setting up a situation whereby subsequently changing the (desired behaviour of) interaction between pinning and non-pinning and settings cannot be designed in such a way to cause a break in the default behaviour on which many people rely -- overall, a test that ensures clarity, rather than saying that the behaviour is provided by the combination of tests which may reasonably change for other reasons.

I am proposing:
   - `testDefaultPolicyFailsToEvaluateServerTrustFromSelfSignedCertificateWithNonMatchingHostname`
   - `testDefaultPolicyFailsToEvaluateServerTrustFromSelfSignedCertificateWithMatchingHostname`
   - `testDefaultPolicyFailsToEvaluateServerTrustFromTrustedCertificateWithNonMatchingHostname`
   - `testDefaultPolicyEvaluatesServerTrustFromTrustedCertificateWithMatchingHostname`

or all together:
   - `testDefaultPolicyMatchesTrustedCertificateWithMatchingHostnameAndRejectsOthers`

~

I will write a patch for the all-in-one test in the next day-or-so (started but not finished), but please share any other thoughts on this, for me the end goal is simply 'AFNetworking behaves in a reliable and predictable manner in relation to security and the developers have minimised the risk of future regressions' and the above is hopefully a step further in that direction.

duttski:
Added
testDefaultPolicyMatchesTrustedCertificateWithMatchingHostnameAndRejects
Others which incorporates several tests to ensure the default mode of
security is at an expected level

Fixes #2633 

Please review!

### DIFF
--- a/Tests/Tests/AFSecurityPolicyTests.m
+++ b/Tests/Tests/AFSecurityPolicyTests.m
@@ -514,11 +514,35 @@ static SecTrustRef AFUTTrustWithCertificate(SecCertificateRef certificate) {
     XCTAssert(policy.SSLPinningMode==AFSSLPinningModeNone, @"Default policy is not set to AFSSLPinningModeNone.");
 }
 
-- (void)testDefaultPolicyFailsToEvaluateServerTrustFromSelfSignedCertificate {
-    AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
-    SecTrustRef trust = AFUTTrustWithCertificate(AFUTSelfSignedCertificateWithoutDomain());
-    XCTAssert([policy evaluateServerTrust:trust forDomain:nil] == NO, @"Unmatched certificate and pinning mode None should fail");
-    CFRelease(trust);
+- (void)testDefaultPolicyMatchesTrustedCertificateWithMatchingHostnameAndRejectsOthers {
+    {
+        //check non-trusted certificate, incorrect domain
+        AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
+        SecTrustRef trust = AFUTTrustWithCertificate(AFUTSelfSignedCertificateWithCommonNameDomain());
+        XCTAssert([policy evaluateServerTrust:trust forDomain:@"different.foobar.com"] == NO, @"Invalid certificate with mismatching domain should fail");
+        CFRelease(trust);
+    }
+    {
+        //check non-trusted certificate, correct domain
+        AFSecurityPolicy *policy = [AFSecurityPolicy defaultPolicy];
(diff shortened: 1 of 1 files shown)


mattt:
This looks good to me. Thanks so much for your work on this, @duttski. I really appreciate it.

PR #2639, opened 2015-04-05
Update AFURLResponseSerialization.m

shicang1990:
Add content-type:@"text/html"


shicang1990:
My project get data by server,but server returen data content-type is text/html that  AFHTTPRequestOperation GET or POST method go to the failure block

mattt:
`text/html` is not a MIME type for JSON. Fix your server or set this on your own `acceptableContentTypes` property.

shicang1990:
thank you

PR #2666, opened 2015-04-19
Removing unnecessary code in UIImageView+AFNetworking.

imaks:
There is no reason to set `af_imageRequestOperation` to nil as it was already nilled out in [cancelImageRequestOperation](https://github.com/imaks/AFNetworking/blob/remove-unnecessary-code/UIKit%2BAFNetworking/UIImageView%2BAFNetworking.m#L128).


mattt:
This assumption does not hold when calling this method from multiple threads. Unless this fixes a specific problem, I'm not particularly interested in integrating it, for risk of introducing a regression.

PR #2675, opened 2015-04-22
Added new constructor to AFURLRequestSerialization

bw0lff:
Added new constructor to AFURLRequestSerialization enabling it to use a custom NSStringEncoding.
AFURLRequestSerialization already supports variable NSStringEncoding, but there was no way to set it...


mattt:
`AFHTTPRequestSerializer` [does indeed have a `stringEncoding` property](https://github.com/bw0lff/AFNetworking/blob/11a463d10487262061d458d192d7fe3044dbd1d4/AFNetworking/AFURLRequestSerialization.h#L68-L71). No additional constructor is necessary.

PR #2683, opened 2015-04-24
Fix documentation of AFSecurityPolicy

tamastimar:
according to 3e631b2.


mattt:
:+1:

Thanks, @tamastimar.

PR #2686, opened 2015-04-26
Faster JSON NULL removal

dtorres:
By initially creating the containers as mutable, the object creation is now 1 time for best case scenario (when consumer wants mutable containers) and 2 for the worst (when the consumer wants immutable containers), as opposed to before where the cases were 2 and 3 respectively.

Also added a performance test for this.

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -57,26 +57,38 @@ static BOOL AFErrorOrUnderlyingErrorHasCodeInDomain(NSError *error, NSInteger co
     return NO;
 }
 
-static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingOptions readingOptions) {
+static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, BOOL mutableContainers) {
     if ([JSONObject isKindOfClass:[NSArray class]]) {
-        NSMutableArray *mutableArray = [NSMutableArray arrayWithCapacity:[(NSArray *)JSONObject count]];
-        for (id value in (NSArray *)JSONObject) {
-            [mutableArray addObject:AFJSONObjectByRemovingKeysWithNullValues(value, readingOptions)];
+        NSMutableArray *mutableArray = JSONObject;
+        NSArray *immutableArray = [JSONObject copy];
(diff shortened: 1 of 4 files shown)


phoney:
I think that the best case would be 0 object creation. If there are no NSNull nulls and the user wants immutable objects.

kcharwood:
Test Results for current master:

```
Test Case '-[AFJSONResponseSerializationTests testResponseSerializerPerformance]' started.
/Users/kevin.harwood/GIT/AFNetworking/Tests/Tests/AFJSONSerializationTests.m:146: Test Case '-[AFJSONResponseSerializationTests testResponseSerializerPerformance]' measured [Time, seconds] average: 1.310, relative standard deviation: 1.383%, values: [1.344652, 1.301903, 1.293353, 1.300250, 1.302915, 1.300436, 1.310927, 1.295425, 1.344801, 1.301231], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
```

Test Results with this patch:
```
Test Case '-[AFJSONResponseSerializationTests testResponseSerializerPerformance]' started.
/Users/kevin.harwood/GIT/AFNetworking/Tests/Tests/AFJSONSerializationTests.m:146: Test Case '-[AFJSONResponseSerializationTests testResponseSerializerPerformance]' measured [Time, seconds] average: 0.740, relative standard deviation: 4.882%, values: [0.748341, 0.700258, 0.707570, 0.750609, 0.693953, 0.805468, 0.788502, 0.755892, 0.741641, 0.705121], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.10
```

kcharwood:
Thanks for the PR. I'm going to leave this out for now erring on the side of reducing churn in this code path since _everyone_ uses JSON parsing.

If we ever see issues with parsing performance as a significant bottleneck for users, we'll definitely revisit this PR.

Thanks! :beers:

PR #2687, opened 2015-04-27
fix bug that when outputstream is unavailable should break the While recycle and return error

jianpx:
fix bug that when outputstream is unavailable should break the While recycle and return error


kcharwood:
Thanks for the PR @jianpx! Would you be willing to create some tests around this?

:beers:

kcharwood:
This was fixed in #2496

PR #2688, opened 2015-04-28
Minor changes to support modern Objective-C syntax.

mattshedlick:
1. NSDictionary  +dictionaryWithObject:obj forKey:key -> @{key : obj}
2. NSDictionary -objectForKey:key -> dict[key]
3. NSArray -objectForKey:0 -> [array firstObject] (in AFURLRequestSerialization.m where it makes sense).
4.NSArray +arrayWithObject[x] -> @[x]

### DIFF
--- a/AFNetworking/AFURLConnectionOperation.m
+++ b/AFNetworking/AFURLConnectionOperation.m
@@ -499,7 +499,7 @@ static inline BOOL AFStateTransitionIsValid(AFOperationState fromState, AFOperat
 - (void)cancelConnection {
     NSDictionary *userInfo = nil;
     if ([self.request URL]) {
-        userInfo = [NSDictionary dictionaryWithObject:[self.request URL] forKey:NSURLErrorFailingURLErrorKey];
+        userInfo = @{NSURLErrorFailingURLErrorKey : [self.request URL]};
     }
     NSError *error = [NSError errorWithDomain:NSURLErrorDomain code:NSURLErrorCancelled userInfo:userInfo];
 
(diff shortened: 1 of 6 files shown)


kcharwood:
We should actually be looking into moving everything to modern syntax... Perhaps this PR could take that on?

tewha:
I don't think a third party can help with a pull request, can we? It might be better to land this (assuming it's good, I haven't looked) and ask in another issue. This is something I'd be happy to do if someone doesn't get to it first.

kcharwood:
@tewha You can add his [remote](https://github.com/mattshedlick/AFNetworking/tree/modern_syntax) to your local git instance, and branch off his commit and continue. You could then push up a new PR and his commit history will still be included.

mattshedlick:
I think I got all the object subscripting changes in this PR.

kcharwood:
@mattshedlick cool. I havent given it a thorough review, but will do so now knowing you think you got them all. I'll try and do an audit as well.

kcharwood:
I ran the modern objective-c conversion tool, and ended up with this commit:

https://github.com/AFNetworking/AFNetworking/commit/ff029cf913a0efc08f9fba93574ecf05bea600b2

Any thoughts on that?

tewha:
I noticed an error from the upgrade two and commented on it; it changed an action to a read only property.

kcharwood:
I've made those small updates called out on that commit.

It looks like the conversion tool added `NS_DESIGNATED_INITIALIZER` in the wrong places as well, causing a few compile warnings. I'll clean those up.

kcharwood:
Alright, 67fb333 now has been squashed down to contain all the changes on top of @mattshedlick's commit.

kcharwood:
All tests pass locally.

Thanks for the pr!

:beers:

Issue #2566, opened 2015-02-17
Adding "nullability" annotations?


dapenggao:
The new Clang compiler allows Objective-C code to have "nullability annotations." Should we add them into the project so that it can interact with Swift code better?

franklsf95:
+1 would love to see this.

kcharwood:
I'd be happy to review this if anyone wants to take a stab at the initial implementation.

franklsf95:
I would love to work on this. Progress at https://github.com/franklsf95/AFNetworking.

kcharwood:
Sounds great. Feel free to create a PR when its ready for review!

PR #2706, opened 2015-05-07
Adding nullability annotations

franklsf95:
I have completed preliminarily adding `nonnull` and `nullable` annotations to the current framework. I would love anybody interested in this to review the header files. Something specific to notice:

- `FIXME`s: there are a few places that I'm not 100% sure whether the argument should be nullable. They are marked with `FIXME`.
- Deprecated methods are not annotated.
- Please double-check all the **block** parameter types.
- Initializers are **not** annotated as `nullable` although they indeed might return `nil`. I'm not sure what the best practice/convention is.

Feel free to suggest changes and I'll be happy to prepare another PR for merge.

### DIFF
--- a/AFNetworking/AFHTTPRequestOperation.h
+++ b/AFNetworking/AFHTTPRequestOperation.h
@@ -22,6 +22,8 @@
 #import <Foundation/Foundation.h>
 #import "AFURLConnectionOperation.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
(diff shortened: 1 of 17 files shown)


kcharwood:
I'm guessing this is failing because Travis isnt running Xcode 6.3

kcharwood:
Unfortunately Travis doesn't have Xcode 6.3 yet...

At the moment, ensuring the test suite runs on every PR is more important than getting this in. I'll leave this open until we can run it on CI.

segiddins:
Why not add conditional defines for the nullability annotations so this can also work in 6.2?

kcharwood:
Linking with #2566 

Thanks so much @franklsf95 

I just got around to reviewing it all. Made a few minor changes and addressed your comments you left. Will get this merged in today.

:beers:

Issue #2660, opened 2015-04-16
AFNetworkActivityIndicatorManager doesn't work for background downloads


phoney:
I'm using AFHTTPSessionManager and I create it with a background NSURLSessionConfiguration. When I do a GET the networkActivityIndicator doesn't appear in the status bar. In the app I'm working on if I use the identical code but use a default SessionConfiguration the networkActivityIndicator works as expected.

This previously did work. In the past the networkActivityIndicator didn't work at all for iOS 7.x but did work for iOS 8 for default and background downloads. Now default downloads show the networkActivityIndicator on iOS 7.x and 8.x but when using background downloads the networkActivityIndicator doesn't appear on iOS 7.x and 8.x.

I believe the problem lies with the method swizzling code in AFURLSessionManager, which changed recently. I notice that for background downloads when I send resume to the NSURLSessionDownloadTask the af_resume method isn't called, like it is for non-background downloads. Since the AFNetworkActivityIndicatorManager is dependent on the resume method causing a notification to be posted the networkActivityIndicator doesn't appear.

BTW, the first two unit tests in AFNetworkActivityManagerTests.m are faulty. The completion blocks are never executed so the expect() blocks inside the completion blocks are never executed either. Regardless, there is no test for the scenario that I describe. 

Sorry that I don't have a simple test case to show you. I did spend some time trying to build a simple test based on the code in AFNetworkActivityManagerTests.m but I mostly just got crashes. A valid test would simply create an AFURLSessionManager with a background session configuration. Then start a download and test that the networkActivityIndicator is visible. I think you'll find that it isn't visible.

phoney:
I've written a small test app to demonstrate this problem. You can download it from https://dl.dropboxusercontent.com/u/17892523/NoProgress.zip

kcharwood:
This should be fixed when #2702 gets merged in. 

Thanks for work in reviewing that PR @phoney

PR #2702, opened 2015-05-08
_AFStateObserving Swizzling Implementation is Wrong

kcharwood:
The `_AFStateObserving` category on `NSURLSessionTask` was recently changed (2.5.3) to remove a `dispatch_once` block.  I'm not sure what the intent of this change was, but one of the side effects is that the swizzling gets run multiple times (for every initialized subclass of `NSURLSessionTask`) and effectively "flip flops" the implementations of the original methods with the swizzled methods.

If this gets called an odd number of times, it happens to work.  However, if it gets called an even number of times, the net result is that the swizzled methods are never run.

If I log the classes this method is called on, here's the result I see when I first use a data task:

```
2015-05-06 14:17:00.788 class=__NSCFLocalDataTask, taskClass=__NSCFLocalSessionTask
2015-05-06 14:17:00.793 class=__NSCFLocalSessionTask, taskClass=__NSCFLocalSessionTask
2015-05-06 14:17:00.795 class=NSURLSessionTask, taskClass=__NSCFLocalSessionTask
2015-05-06 14:17:00.805 class=__NSCFLocalUploadTask, taskClass=__NSCFLocalSessionTask
2015-05-06 14:17:00.809 class=NSURLSessionDataTask, taskClass=__NSCFLocalSessionTask
```

So, this works fine for a while.  However, if I eventually _also_ use an upload task, I see this also:

```
2015-05-06 14:19:51.903 class=NSURLSessionUploadTask, taskClass=__NSCFLocalSessionTask
```

This last call swizzles the methods a sixth time, resulting in `resume`/`suspend` being set back to their original implementations.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -255,68 +255,121 @@ expectedTotalBytes:(int64_t)expectedTotalBytes {
 
 #pragma mark -
 
-/*
- A workaround for issues related to key-value observing the `state` of an `NSURLSessionTask`.
-
- See https://github.com/AFNetworking/AFNetworking/issues/1477
+/**
+ *  A workaround for issues related to key-value observing the `state` of an `NSURLSessionTask`.
+ *
+ *  See:
+ *  - https://github.com/AFNetworking/AFNetworking/issues/1477
+ *  - https://github.com/AFNetworking/AFNetworking/issues/2638
+ *  - https://github.com/AFNetworking/AFNetworking/pull/2702
  */
 
 static inline void af_swizzleSelector(Class class, SEL originalSelector, SEL swizzledSelector) {
     Method originalMethod = class_getInstanceMethod(class, originalSelector);
     Method swizzledMethod = class_getInstanceMethod(class, swizzledSelector);
-    if (class_addMethod(class, originalSelector, method_getImplementation(swizzledMethod), method_getTypeEncoding(swizzledMethod))) {
-        class_replaceMethod(class, swizzledSelector, method_getImplementation(originalMethod), method_getTypeEncoding(originalMethod));
-    } else {
-        method_exchangeImplementations(originalMethod, swizzledMethod);
-    }
+    method_exchangeImplementations(originalMethod, swizzledMethod);
 }
 
-static inline void af_addMethod(Class class, SEL selector, Method method) {
-    class_addMethod(class, selector,  method_getImplementation(method),  method_getTypeEncoding(method));
+static inline BOOL af_addMethod(Class class, SEL selector, Method method) {
+    return class_addMethod(class, selector,  method_getImplementation(method),  method_getTypeEncoding(method));
 }
 
 static NSString * const AFNSURLSessionTaskDidResumeNotification  = @"com.alamofire.networking.nsurlsessiontask.resume";
 static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofire.networking.nsurlsessiontask.suspend";
 
-@interface NSURLSessionTask (_AFStateObserving)
+@interface _AFURLSessionTaskSwizzling : NSObject
+
 @end
 
-@implementation NSURLSessionTask (_AFStateObserving)
+@implementation _AFURLSessionTaskSwizzling
 
-+ (void)initialize {
-    if ([NSURLSessionTask class]) {
-        NSURLSessionDataTask *dataTask = [[NSURLSession sessionWithConfiguration:nil] dataTaskWithURL:nil];
-        Class taskClass = [dataTask superclass];
++ (void)load {
+    /**
+     WARNING: Trouble Ahead
+     https://github.com/AFNetworking/AFNetworking/pull/2702
+     */
 
-        af_addMethod(taskClass, @selector(af_resume),  class_getInstanceMethod(self, @selector(af_resume)));
-        af_addMethod(taskClass, @selector(af_suspend), class_getInstanceMethod(self, @selector(af_suspend)));
-        af_swizzleSelector(taskClass, @selector(resume), @selector(af_resume));
-        af_swizzleSelector(taskClass, @selector(suspend), @selector(af_suspend));
-
-        [dataTask cancel];
+    if (NSClassFromString(@"NSURLSessionTask")) {
+        /**
+         iOS 7 and iOS 8 differ in NSURLSessionTask implementation, which makes the next bit of code a bit tricky.
+         Many Unit Tests have been built to validate as much of this behavior has possible.
+         Here is what we know:
+            - NSURLSessionTasks are implemented with class clusters, meaning the class you request from the API isn't actually the type of class you will get back.
+            - Simply referencing `[NSURLSessionTask class]` will not work. You need to ask an `NSURLSession` to actually create an object, and grab the class from there.
+            - On iOS 7, `localDataTask` is a `__NSCFLocalDataTask`, which inherits from `__NSCFLocalSessionTask`, which inherits from `__NSCFURLSessionTask`.
+            - On iOS 8, `localDataTask` is a `__NSCFLocalDataTask`, which inherits from `__NSCFLocalSessionTask`, which inherits from `NSURLSessionTask`.
+            - On iOS 7, `__NSCFLocalSessionTask` and `__NSCFURLSessionTask` are the only two classes that have their own implementations of `resume` and `suspend`, and `__NSCFLocalSessionTask` DOES NOT CALL SUPER. This means both classes need to be swizzled.
+            - On iOS 8, `NSURLSessionTask` is the only class that implements `resume` and `suspend`. This means this is the only class that needs to be swizzled.
+            - Because `NSURLSessionTask` is not involved in the class hierarchy for every version of iOS, its easier to add the swizzled methods to a dummy class and manage them there.
+        
+         Some Assumptions:
+            - No implementations of `resume` or `suspend` call super. If this were to change in a future version of iOS, we'd need to handle it.
+            - No background task classes override `resume` or `suspend`
+         
+         The current solution:
+            1) Grab an instance of `__NSCFLocalDataTask` by asking an instance of `NSURLSession` for a data task.
+            2) Grab a pointer to the original implementation of `af_resume`
+            3) Check to see if the current class has an implementation of resume. If so, continue to step 4.
+            4) Grab the super class of the current class.
+            5) Grab a pointer for the current class to the current implementation of `resume`.
+            6) Grab a pointer for the super class to the current implementation of `resume`.
+            7) If the current class implementation of `resume` is not equal to the super class implementation of `resume` AND the current implementation of `resume` is not equal to the original implementation of `af_resume`, THEN swizzle the methods
+            8) Set the current class to the super class, and repeat steps 3-8
+         */
+        NSURLSessionDataTask *localDataTask = [[NSURLSession sessionWithConfiguration:nil] dataTaskWithURL:nil];
+        IMP originalAFResumeIMP = method_getImplementation(class_getInstanceMethod([_AFURLSessionTaskSwizzling class], @selector(af_resume)));
+        Class currentClass = [localDataTask class];
+        
+        while (class_getInstanceMethod(currentClass, @selector(resume))) {
+            Class superClass = [currentClass superclass];
+            IMP classResumeIMP = method_getImplementation(class_getInstanceMethod(currentClass, @selector(resume)));
+            IMP superclassResumeIMP = method_getImplementation(class_getInstanceMethod(superClass, @selector(resume)));
+            if (classResumeIMP != superclassResumeIMP &&
+                originalAFResumeIMP != classResumeIMP) {
+                [self swizzleResumeAndSuspendMethodForClass:currentClass];
+            }
+            currentClass = [currentClass superclass];
+        }
+        
+        [localDataTask cancel];
     }
 }
 
-#pragma mark -
++ (void)swizzleResumeAndSuspendMethodForClass:(Class)class {
+    Method afResumeMethod = class_getInstanceMethod(self, @selector(af_resume));
+    Method afSuspendMethod = class_getInstanceMethod(self, @selector(af_suspend));
+    
+    af_addMethod(class, @selector(af_resume), afResumeMethod);
+    af_addMethod(class, @selector(af_suspend), afSuspendMethod);
+    
+    af_swizzleSelector(class, @selector(resume), @selector(af_resume));
+    af_swizzleSelector(class, @selector(suspend), @selector(af_suspend));
+}
+
+- (NSURLSessionTaskState)state {
+    NSAssert(NO, @"State method should never be called in the actual dummy class");
+    return NSURLSessionTaskStateCanceling;
+}
 
 - (void)af_resume {
-    NSURLSessionTaskState state = self.state;
+    NSAssert([self respondsToSelector:@selector(state)], @"Does not respond to state");
+    NSURLSessionTaskState state = [self state];
     [self af_resume];
-
+    
     if (state != NSURLSessionTaskStateRunning) {
         [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidResumeNotification object:self];
     }
 }
 
 - (void)af_suspend {
-    NSURLSessionTaskState state = self.state;
+    NSAssert([self respondsToSelector:@selector(state)], @"Does not respond to state");
+    NSURLSessionTaskState state = [self state];
     [self af_suspend];
-
+    
     if (state != NSURLSessionTaskStateSuspended) {
         [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidSuspendNotification object:self];
     }
 }
-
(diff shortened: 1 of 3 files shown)


kcharwood:
It looks like Travis just can't handle the XCTests with expectationsForNotifications. I've commented those out for now, since Travis is running Xcode 6.1.

The tests do work locally for me.

AFNetworking/AFURLSessionManager.m
```diff
@@ -319,6 +303,21 @@ - (void)af_suspend {
...
 @end
 
+@interface _AFURLSessionTaskSwizzling : NSObject
+
+@end
+
+@implementation _AFURLSessionTaskSwizzling
+
++ (void)load {
+    Class urlSessionTaskClass = NSClassFromString(@"NSURLSessionTask");
```
fouvrai: I think this fix ignores the issue that 27964aa9dc8b3f5aece2c1528e8135355ee21156 was ostensibly trying to fix.  I'm not familiar with the history of this code, but presumably swizzling the superclass of `NSURLSessionDataTask` is important, and this change reverts that behavior.

kcharwood: Thanks for the note @fouvrai. It's difficult to track the myriad of issues associated to this swizzling due to the varying implementations that have been in the past few releases. Let me take a look once more with that lens and make sure we are good.

kcharwood:
@fouvrai I think you may be right, as I am seeing some strange behavior on iOS 7. How many different ways can I possibly skin this cat...

kcharwood:
I'm starting to question if this has every actually worked properly for iOS 7. I have an example that works properly for NSURLSessionDataTask, but trying to use UploadTask or DownloadTask causes a problem on iOS 7. It all works great on iOS 8.

Running various variations of the swizzling I've seen in the history of this file seems to fail in the same way. This one is nasty, mainly due to the implementation differences under the hood between iOS 8 and iOS 7 that Apple has hidden away from us.

kcharwood:
I've built out a class diagram for NSURLSession for iOS 7 and iOS 8, based on what classes are being given to me at runtime.

iOS 7:
screen shot 2015-05-08 at 9 31 36 pm

iOS 8:
screen shot 2015-05-08 at 9 31 42 pm

Note the grey boxes, which represent the hierarchy if you do something like `[[NSURLSessionDataTask class] superclass]`. The green boxes represent the class hierarchy if you take an actual object returned by `NSURLSessionManager`, and go up the superclass tree. I have tests confirming the swizzling approach works for all green boxes.

Note the one red box in iOS 7 - The upload task object. Building out this diagram helped me identify that as the one failing test. If I disable the tests for UploadTask on iOS 7, everything else passes. If I run the UploadTask tests in iOS 7 in isolation or in the suite, it causes a deadlock inside of `af_suspend`.

I have not been able to fully diagnose why this is happening, but I'm using this ticket as a scratchpad to talk out my notes.

larsacus:
I know I might be commenting on this without the full history or context as to what's going on with this particular bit of code, but when someone says 'runtime', my ears perk up.

So to summarize, it appears that the entire purpose of the swizzled methods is to: 

1. Add the `af_resume`/`af_suspend` methods to all (what we think is) the root class for _all_ `NSURLSessionTask` objects used by the user
2. Call those newly added methods on the existing implementations of `-[NSURLSessionTask resume]` and `-[NSURLSessionTask suspend]` in order to post a notification

In addition, the reason that `+initialize` is being used as opposed to `+load` is due to not actually knowing what the true "root" class ("root" here being the closest superclass after `NSObject`) of the hierarchy tree is. This is due to the class hierarchy diagrams that Kevin drew up since `+load` is called in order of hierarchy, and this breaks on iOS 7.

Is there a reason we need to swizzle this into the _class_ of `NSURLSessionTask`, so that the new functionality is to be used on all classes, even those that are created outside of AFNetworking? Could we simply add and swizzle in the same exact methods, but on a per-instance basis in the `*TaskWithRequest` methods in `AFURLSessionManager`, where everything is funneling into? We could then add/swizzle on the actual returned class (once) vs an unknown root class (which would run an unknown number of times in a runtime method like `+initialize`).

Bad? Depends on if this functionality must live on the furthest superclass and if all `*SessionTask` subclasses in the bundle need this functionality regardless of association with AFNetworking.

Another solution could be to actually _use_ the runtime to find the furthest superclass that actually implements suspend/resume and add the `af_*` methods to just that one class. Then use the existence of those `af_*` methods we just added to know if we have already performed this action, or also wrap this action into a `dispatch_once`.

I'm also not positive what is causing the deadlocking referred to above. I'm assuming the `onceToken`, since it is static and only found in that one category, is actually the same token across multiple calls to `+initialize`. This would mean that `+intialize` is called concurrently?

phoney:
Removing the dispatch_once is OK. The normal way that +initialize is protected so that the code is run only once is like this

+(void)initialize
{
   if ([self class] == [NSURLSessionTask class] {
         // code that should only run once goes here
   }
}

I think in the code committed by @mattt https://github.com/AFNetworking/AFNetworking/commit/ace91df295c1ed9800be5d3fc660f99ca5fc8792  he had a typo where the == [NSURLSessionTask class] part was missing.

This entire issue is the cause of my related issue: https://github.com/AFNetworking/AFNetworking/issues/2660 AFNetworkActivityIndicatorManager doesn't work for background downloads.

If you change this code again I request that you make sure it works for background downloads.

I think the swizzling can be made to work for background downloads if you swizzle [self class] one time.

My real opinion is that we should throw away the swizzling code and go back to the KVO code. KVO observing of the task state is the right way to do this. I understand there were some crashes in some cases when removing the observers. I expect that those crashes could be fixed. Swizzling is an attempt to avoid fixing the KVO code. Everyone at Apple says don't swizzle. They're obviously right.

AFNetworking/AFURLSessionManager.m
phoney: Swizzling of the superclass was adding here https://github.com/AFNetworking/AFNetworking/commit/27964aa9dc8b3f5aece2c1528e8135355ee21156 Fix AFNetworkActivityIndicatorManager for iOS 7...dubiously by @tangphillip He explains the reason is due to the superclass being an undocumented class in iOS7.

tangphillip: @phoney has it absolutely right here. This is not going to work in iOS 7, because in the actual implementation of iOS 7, `NSURLSessionDataTask`'s superclass chain doesn't include a class named `NSURLSessionTask`. Instead, its superclass is named `__NSCFLocalSessionTask`. Which is private, and probably not a good idea to reference directly.

tangphillip: These swizzles were (are?) in an `+initialize` of a category on `NSURLSessionDataTask` to guarantee that the entire inheritance chain of `NSURLSession` classes had an opportunity to `+initialize` before we swizzle their methods.

tangphillip:
Those crashes are not fixed—they're a bug in KVO itself..and it's still at large, as of iOS 8.1. The KVO crashes increase our per-session crash rate by ~0.2% (leading to hundreds of crashes a day), making them our most common crash by a factor of 10.

I agree that swizzling is a _terrible_, _terrible_ idea. But ultimately, code quality is less important than user experience, which is why swizzling is unfortunately the right thing to do here.

AFNetworking/AFURLSessionManager.m
kcharwood: @tangphillip yes you are correct. The branch I have pushed here does not match my latest branch. You're right this approach won't work on 7.

I have something right now that works for 8 and everything on 7 except for Upload Task. Haven't had time to try a few of my other ideas yet.

kcharwood:
I pushed up a rebase off master with just the url session unit tests I have for this issue if anyone wants to try their solution on it.

https://github.com/AFNetworking/AFNetworking/tree/updated_url_session_tests

phoney:
@tangphillip Regardless if this issue is due to some defect in KVO, KVO is used by the OS and by thousands and thousands of apps on iOS and MacOS X. Fixing the KVO solution is a better idea than method swizzling. From the stack traces in the previous discussions it seems that this is a thread safety problem related to removing observers from transfers that end in failure. Pushing the removal of observers to a later time, and perhaps to the main thread would be a likely way to fix this.

kcharwood:
OKKKKKKKK.......

So I've got the swizzling option working in iOS 7 and iOS 8. The last two problems I ran into:

* There is a bug on iOS 7 where passing nil data into an upload tasks causes the deadlock. That appears to be fixed under the hood in iOS 8. Adding dummy data to my tests allows the tests to pass on iOS 7
* There was an issue with the activity indicator tests that appeared to allow mocked objects to live on past the conclusions of those tests causing random failures. @cnoon fixed that issue, and now travis is happy...

As far as the KVO vs swizzling debate: I'm by no means a swizzling advocate, but it does have its place. This isn't a terrible option to hook into those methods since we can't subclass, but in light of the KVO issues we have been seeing, it seems like the better approach. Now that we have tests around this, it should help us prevent having regressions in the future.

If anyone takes a stab at a KVO approach, I'll be happy to entertain it, but I'd like to see lots of tests around it.

Anyone have any issues with the PR as is right now?

phoney:
@kcharwood Your pull request code isn't working for me for background downloads. On iOS 8 I get an infinite recursion in af_resume. In iOS 7 I don't see the applicationNetworkActivity indicator spin.

If you look at the NoProgress sample app from my issue https://github.com/AFNetworking/AFNetworking/issues/2660 and add your AFNetworking code to it you should see the infinite recursion. That code doesn't work right on iOS 7 though.

BTW, Apple reserves class names and method names that start with '_' I assume you've noticed the various private classes whose names start with '_'  It's probably not a good idea to name class extensions with names that start with '_'

kcharwood:
@phoney do the tests pass on your machine? Is it possible to build a test demonstrating the problem? I'd to work that way if possible so we can have a reproducible problem.

phoney:
Sure they pass, you have no tests for background transfers. Change to this and you'll see the failure on iOS 8

@implementation AFURLSessionManagerTests

- (void)setUp {
    [super setUp];
//    self.manager = [[AFURLSessionManager alloc] init];
    self.manager = [[AFURLSessionManager alloc] initWithSessionConfiguration:[NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:@"com.alamofire.test"]];
}

kcharwood:
Yep I'm seeing it... and I'm seeing more private classes obscured by the class cluster. Time for more tests...

kcharwood:
@phoney try my latest push to this branch.

I've added a bunch of tests around the swizzling to ensure the map is setup correctly for foreground and background tasks.

Due to the absolute craziness under the hood of URLSession, it looks like I can't be built a background session manager on iOS 7 without throwing an exception and crashing the test target, so I'm getting a little creative at the moment with how I test it.

If it looks good for you, let me know and I'll put some final clean up.

As an aside, I'm still interested in a KVO workaround because at this point this is getting a little out of control, and I'm terrified to see what Apple does with iOS 9 under the hood that breaks all these assumptions as well.

Thanks again for work here.

phoney:
I'll look at the code tomorrow. While I think the KVO code could be made to work by waiting until the state of the task is in a completed value before removing observers, I'm starting to think we should go another another way. 

All we need is notifications for resume, suspend, and didComplete for all the tasks. The NSURLSessionDelegate gets notified for didComplete and already posts the AFNetworkingTaskDidCompleteNotification notification. We can simply add resumeTask: and suspendTask: methods to AFURLSessionManager that send resume and suspend to the task and post the AFNotifications. I understand that this is not quite as automatic as KVO or the swizzling strategies but should be reliable.

Another way to make this slightly nicer would be to add af_suspend and af_resume methods to the NSURLSessionTasks in categories. Those methods would call self suspend/resume and post the notifications. This is almost exactly the same as the swizzling strategy but is much more compatible.

Backwards compatibility is a question here that I haven't thought through completely. In order to use this new functionality any AFURLSessionManager subclasses would need to call the new methods instead of [task resume] or [task suspend]. In fact that would be a trivial change to existing source code but it would break existing code otherwise.

Anyway, what do you think about this? It seems much simpler and much more compatible than swizzling.

tangphillip:
The solution you propose works, except that other people can call resume and suspend on the tasks—we return the tasks from the AFURLSessionManager methods that create them.

kcharwood:
Running rake locally on machine shows all tests passing. Running on Travis shows failures. I can't catch a break... Looks like the failures are for 10.9.5. Trying to find a 10.9 machine now.

@phoney personally I'm not a huge fan of asking everyone to use custom resume/suspend methods, as that can be confusing and as you said, not backwards compatible.  If we can "make it magically work", I'm more of a fan of that approach.

cnoon:
+1 @kcharwood. I think we all thought through alternative possibilities while maintaining backwards compatibility. That's why swizzling is really the only option since KVO is crashing. All other proposed solutions would have serious public API implications or would break this feature for all our current users.

kcharwood:
TRAVIS CI JUST PASSED MY TESTS. I'm going to spend quite a bit of time cleaning this branch up now, and adding appropriate documentation.

kcharwood:
Ok. Branch cleaned up. More tests added. Comments added to explain my thoughts.

I would love as many eyes as possible on this one.

cnoon:
+1 on the change in 2de01f66a39b90dae3d8ce3dbed3bf145bdd509a. It's necessary due to the mocking not being town down properly.

phoney:
I tested your newest code in my app on iOS 8 and iOS 7 and it seems to work fine. The network activity indicator spins for background downloads and non-background requests.

AFNetworking/AFURLSessionManager.m
```diff
@@ -255,68 +255,130 @@ - (void)URLSession:(__unused NSURLSession *)session
...
+    af_swizzleSelector(class, @selector(resume), @selector(af_resume));
+    af_swizzleSelector(class, @selector(suspend), @selector(af_suspend));
+}
 
 - (void)af_resume {
-    NSURLSessionTaskState state = self.state;
+    NSURLSessionTaskState state;
+    SEL selector = @selector(state);
+    NSAssert([self respondsToSelector:selector], @"Does not respond to state");
+    NSInvocation *invocation = [NSInvocation invocationWithMethodSignature:
```
fouvrai: What's the rationale for switching to an `NSInvocation` here?  What's wrong with `NSURLSessionTaskState state = [self state];`?

kcharwood: These methods are in a dummy class, which doesn't have a self state. To prevent a compile warning, I went this route.

I guess it would be possible to stub out the state method in the dummy class and drop an assert in there.

fouvrai: Ah, yea I was only thinking about it in the context in which it'd actually be called.

It seems weird to me to compromise the implementation to prevent a compiler warning, when the resultant implementation wouldn't actually work on the class on which it's defined anyway.  At that point, it feels like it'd be better to just locally disable the compiler warning, rather than simply "hiding" it.

An alternative option would be to derive the `_AFURLSessionTaskSwizzling` class from `NSURLSessionTask`, so that the methods these functions need to call on `self` would exist on the class.

Tests/Tests/AFURLSessionManagerTests.m
```diff
@@ -78,4 +102,266 @@ - (void)testDownloadTasksProgressBecomesPartOfCurrentProgress {
...
+                                                               completionHandler:nil];
+        [self _testSuspendNotificationForTask:task];
+    }
+}
+
+- (void)testDidResumeNotificationIsReceivedByLocalUploadTaskAfterResume {
+    NSURLSessionUploadTask *task = [self.localManager uploadTaskWithRequest:[self _delayURLRequest]
+                                                              fromData:[NSData data]
+                                                              progress:nil
+                                                     completionHandler:nil];
```
cnoon: Looks like the code in about half the tests got out of alignment.

cnoon:
These changes all look good to me @kcharwood. I went through it with a fine tooth comb and it looks great! All the documentation you've added is really helpful, and the test coverage is really well done. Awesome job!

AFNetworking/AFURLSessionManager.m
kcharwood: I decided just to stub the method state method in the dummy class. Due to the complications around the class cluster of `NSURLSessionTask`, I'd prefer to keep the dummy class completely out of the inheritance chain.

Removing the NSInvocation does make it a bit cleaner, so I'm happy with compromise.

Tests/Tests/AFURLSessionManagerTests.m
kcharwood: Those are offset because they are offset by that if statement to check for the background manager. The wonky iOS 7 exception that is thrown in the test target by the background manager means the I kind of have to have the tests fake pass on iOS 7, which isnt ideal, but I at least get coverage on iOS 8.

cnoon:
Looks good!

kcharwood:
I'd like to thank everyone who jumped in on this issue and provided feedback/suggestions. Super nasty issue, and couldn't have gotten to this solution without everyones involvement.

cheers :beers:

julianL0veios:
screen shot 2017-04-21 at 11 44 11 am

julianL0veios:
af3.1.0version iOS10.2 still got the crash!
410bbe63-0373-4b84-957e-13cdb016dc00

PR #2717, opened 2015-05-12
Fix Tests Randomly Failing for Reachability and Activity Manager

kcharwood:
Lets get this cleaned up so other branches quit getting false positives.


cnoon:
+1...for some reason the mocking wasn't being torn down properly.

Issue #2673, opened 2015-04-21
AFSSLPinningMode documentation insufficient


tewha:
This could be seen as a question, but I think of it more as a request for documentation improvement. :)

Reading the documentation [here on AFSSLPinningMode](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFSecurityPolicy.h#L124), I think this is vague and there's room for improvement. There could be detail in each mode about what it does in the case of updated certificates, for instance. If I update the certificate on the server, will I need to update the public certificate in the app before it works again?

I'd take a swing at this with a pull request, but I don't really understand the modes myself.

kcharwood:
This would definitely be helpful.

High level: Pinning against the public key does not require you to update the cert bundled with your app. If you pin against the certificate, then you'll need to update the cert in your app bundle as well.

Pinning against the cert is "stronger" since it requires an exact match, but also isn't ideal where you can't control whether an end user has updated to the latest version of your app when you do update your certs. Public Key is a good middle ground for most mobile apps.

tewha:
That's useful info! Is there anything else that you think should be in there?

kcharwood:
We could actually use a full "AFNetworking + Security" guide and go deep on all of the advanced options, and explain why you would want to use each.

bdesham:
This is especially important in light of [this serious vulnerability](http://arstechnica.com/security/2015/04/24/critical-https-bug-may-open-25000-ios-apps-to-eavesdropping-attacks/), which as far as I can tell was mostly the result of poor defaults and poor documentation.

danielnordh:
Not completely related, but still about SSL and security it would be helpful if improved security documentation made clear how to know whether AFNetworking (AFAmazonS3Manager) is using SSL when communicating with S3.

This in light of recent emails from Amazon discontinuing support for SSLv3, which I'm not yet clear requires any change on our part to avoid disruption to S3 communication;

"As of 12:00 AM PDT May 20, 2015, AWS will discontinue support of SSLv3 for securing connections to S3 buckets. Security research published late last year demonstrated that SSLv3 contained weaknesses in its ability to protect and secure communications. These weaknesses have been addressed in Transport Layer Security (TLS), which is the replacement for SSL. Consistent with our top priority to protect AWS customers, AWS will only support versions of the more modern TLS rather than SSLv3."

chiahsien:
Can anyone provide a sample snippet about using public key?

[...]

PR #2733, opened 2015-05-18
Implemented Public Key Hash Pinning

### LINKED ISSUE #2695: SSL Public Key Hash Pinning
I would like to add [SPKI Hash Pinning](https://www.owasp.org/index.php/Certificate_and_Public_Key_Pinning#Hashing) rather than extracting the SPKI from the .cer file in my target. [This is considered best practice by OWASP](https://www.owasp.org/index.php/Certificate_and_Public_Key_Pinning#Hashing) and also how Pinning is implemented in [Chrome, Chromium](http://scarybeastsecurity.blogspot.de/2011/04/fiddling-with-chromiums-new-certificate.html), and [OkHttp](http://square.github.io/okhttp/javadoc/com/squareup/okhttp/CertificatePinner.html). It provides the following benefits:

1. Anonymization of the Public Key: Can be important if the app is concerned about disclosing information during recompilation
2. Convenience: It is more convenient to hard code the hashes in the application than to use .cer files. This allows more easily to change the hashes based on compile flags

I would also like the option of providing a __backup key__ in case the primary key is compromised/lost. This is complimentary to SPKI Hash Pinning as hashing ensures that attackers cannot see the backup SPKI prior to its use. This is also part of [Google's IETF websec-key-pinning draft](https://tools.ietf.org/html/draft-ietf-websec-key-pinning-21#section-4.3).

It would be important to me that a PR makes it into the master branch. That's why I'd like to start the discussion with you now. Are you generally cool with adding this in case the AFSecurityPolicy remains backwards compatible?

arein:
This closes #2695 and #2673 

My implementation adds an easy way to use hashes of the Subject Public Key Info (SPKI) of certificates. It also adds a way to accept backup key(s) in case the primary key gets compromised/lost:

``` objectivec
AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKeyHash];
[policy setValidatesCertificateChain:NO]; // This allows to use backup key(s)
[policy setPinnedPublicKeyHashes:[NSArray arrayWithObjects:@"G2bwK+Vdcc6RPUi2h2A9mOPezwc=", @"backupkeyhash", nil]];
```

The hashes can be retrieved like this. Let's assume that we want to pin google.com:

``` bash
openssl s_client -showcerts -connect google.com:443 # Get google.com Certificate Chain
# Create a file called "google.com.crt" from the leaf certificate
openssl x509 -inform PEM -pubkey -noout -in google.com.crt | openssl pkey -pubin -outform DER | openssl dgst -sha1 -binary | openssl enc -base64
```

The main implementation challenge was that extracting a public key using Cocoa's security library returns a subset of the information that you would usually find in an ASN.1 encoded public key. Luckily, I found [this](http://blog.wingsofhermes.org/?p=42) blog post that outlines how to pad the public key.

### DIFF
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -416,11 +416,15 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
             return trustedPublicKeyCount > 0 && ((self.validatesCertificateChain && trustedPublicKeyCount == [serverCertificates count]) || (!self.validatesCertificateChain && trustedPublicKeyCount >= 1));
         }
         case AFSSLPinningModePublicKeyHash: {
-            NSUInteger trustedPublicKeyHashCount = 0;
-            NSArray *publicKeys = AFPublicKeyTrustChainForServerTrust(serverTrust); // Get Remote Public Keys
-            
             if (self.pinnedPublicKeyHashes == nil) return NO; // No Public Key Accepted
             
+            NSArray *publicKeys = AFPublicKeyTrustChainForServerTrust(serverTrust); // Get Remote Public Keys
+            
+            if (!self.validatesCertificateChain && [publicKeys count] > 0) {
+                publicKeys = @[[publicKeys firstObject]];
+            }
+            
+            NSUInteger trustedPublicKeyHashCount = 0;
             for (id trustChainPublicKey in publicKeys) {
                 SecKeyRef pubKey = (__bridge SecKeyRef)trustChainPublicKey;
                 NSData *keyData = AFSecKeyGetData(pubKey);
@@ -433,7 +437,7 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
                 // Create Hash
                 unsigned int outputLength = CC_SHA1_DIGEST_LENGTH;
                 unsigned char output[outputLength];
-                CC_SHA1(x509.bytes, x509.length, output);
+                CC_SHA1(x509.bytes, (unsigned int) x509.length, output);
                 NSData *hash = [NSMutableData dataWithBytes:output length:outputLength];
                 
                 // Encode the hash in Base64
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -22,12 +22,37 @@
 #import "AFSecurityPolicy.h"
 
 #import <AssertMacros.h>
+#import <CommonCrypto/CommonDigest.h>
 
-#if !defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
 static NSData * AFSecKeyGetData(SecKeyRef key) {
+#if !defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
+    SecItemImportExportKeyParameters params;
+    CFMutableArrayRef keyUsage
+    = (__bridge CFMutableArrayRef) [ NSMutableArray
+                                    arrayWithObjects: (__bridge id)(kSecAttrCanEncrypt), kSecAttrCanDecrypt, nil ];
+    CFMutableArrayRef keyAttributes
+    = (__bridge CFMutableArrayRef) [ NSMutableArray array ];
+    SecExternalFormat format = kSecFormatUnknown;
+    CFDataRef keyData;
+    OSStatus oserr;
+    int flags = 0;
+    
+    memset(&params, 0, sizeof(params));
+    params.version = SEC_KEY_IMPORT_EXPORT_PARAMS_VERSION;
+    params.keyUsage = keyUsage;
+    params.keyAttributes = keyAttributes;
+    
+    oserr = SecItemExport(key, format, flags, &params, &keyData);
+    if (oserr) {
+        fprintf(stderr, "SecItemExport failed\n", oserr);
+        return nil;
+    }
+    
+    return (__bridge NSData *) keyData;
+    
     CFDataRef data = NULL;
 
-    __Require_noErr_Quiet(SecItemExport(key, kSecFormatUnknown, kSecItemPemArmour, NULL, &data), _out);
+    __Require_noErr_Quiet(SecItemExport(key, kSecFormatOpenSSL, kSecItemPemArmour, NULL, &data), _out);
 
     return (__bridge_transfer NSData *)data;
 
@@ -37,8 +62,100 @@ _out:
     }
 
     return nil;
-}
+#else
+    static const uint8_t publicKeyIdentifier[] = "com.afnetworking.publickey";
+    NSData *publicTag = [[NSData alloc] initWithBytes:publicKeyIdentifier length:sizeof(publicKeyIdentifier)];
+    
+    OSStatus sanityCheck = noErr;
+    NSData * publicKeyBits = nil;
+    
+    NSMutableDictionary * queryPublicKey = [[NSMutableDictionary alloc] init];
+    [queryPublicKey setObject:(__bridge id)kSecClassKey forKey:(__bridge id)kSecClass];
+    [queryPublicKey setObject:publicTag forKey:(__bridge id)kSecAttrApplicationTag];
+    [queryPublicKey setObject:(__bridge id)kSecAttrKeyTypeRSA forKey:(__bridge id)kSecAttrKeyType];
+    
+    // Temporarily add key to the Keychain, return as data:
+    NSMutableDictionary * attributes = [queryPublicKey mutableCopy];
+    [attributes setObject:(__bridge id)key forKey:(__bridge id)kSecValueRef];
+    [attributes setObject:@YES forKey:(__bridge id)kSecReturnData];
+    CFTypeRef result;
+    sanityCheck = SecItemAdd((__bridge CFDictionaryRef) attributes, &result);
+    if (sanityCheck == errSecSuccess) {
+        publicKeyBits = CFBridgingRelease(result);
+        
+        // Remove from Keychain again:
+        (void)SecItemDelete((__bridge CFDictionaryRef) queryPublicKey);
+    }
+    
+    return publicKeyBits;
 #endif
+}
+
+// From https://github.com/reference/OpenSSLRSAWrapper/blob/master/OpenSSLRSAWrapper/OpenSSLRSAWrapper.m
+// Helper function for ASN.1 encoding
+static size_t encodeLength(unsigned char * buf, size_t length) {
+    
+    // encode length in ASN.1 DER format
+    if (length < 128) {
+        buf[0] =  (unsigned char) length;
+        return 1;
+    }
+    
+    size_t i = (length / 256) + 1;
+    buf[0] =  (unsigned char) i + 0x80;
+    for (size_t j = 0 ; j < i; ++j) {
+        buf[i - j] = length & 0xFF;
+        length = length >> 8;
+    }
+    
+    return i + 1;
+}
+
+static NSData * AFSecKeyPadX509(NSData * publicKeyBits) {
+    // OK - that gives us the "BITSTRING component of a full DER
+    // encoded RSA public key - we now need to build the rest
+    
+    static const unsigned char _encodedRSAEncryptionOID[15] = {
+        /* Sequence of length 0xd made up of OID followed by NULL */
+        0x30, 0x0d, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86,
+        0xf7, 0x0d, 0x01, 0x01, 0x01, 0x05, 0x00
+    };
+    
+    unsigned char builder[15];
+    NSMutableData * encKey = [[NSMutableData alloc] init];
+    unsigned long bitstringEncLength;
+    
+    // When we get to the bitstring - how will we encode it?
+    if  ([publicKeyBits length ] + 1  < 128 )
+        bitstringEncLength = 1 ;
+        else
+            bitstringEncLength = (([publicKeyBits length ] +1 ) / 256 ) + 2 ;
+            
+            // Overall we have a sequence of a certain length
+            builder[0] = 0x30;    // ASN.1 encoding representing a SEQUENCE
+            // Build up overall size made up of -
+            // size of OID + size of bitstring encoding + size of actual key
+            size_t i = sizeof(_encodedRSAEncryptionOID) + 2 + bitstringEncLength +
+            [publicKeyBits length];
+            size_t j = encodeLength(&builder[1], i);
+            [encKey appendBytes:builder length:j +1];
+    
+    // First part of the sequence is the OID
+    [encKey appendBytes:_encodedRSAEncryptionOID
+                 length:sizeof(_encodedRSAEncryptionOID)];
+    
+    // Now add the bitstring
+    builder[0] = 0x03;
+    j = encodeLength(&builder[1], [publicKeyBits length] + 1);
+    builder[j+1] = 0x00;
+    [encKey appendBytes:builder length:j + 2];
+    
+    // Now the actual key
+    [encKey appendData:publicKeyBits];
+    
+    return encKey;
+}
+
 
 static BOOL AFSecKeyIsEqualToKey(SecKeyRef key1, SecKeyRef key2) {
 #if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
@@ -220,6 +337,10 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
     }
 }
 
+- (void)setPublicKeyHashes:(NSArray *)publicKeyHashes {
+    self.pinnedPublicKeyHashes = publicKeyHashes;
+}
+
 #pragma mark -
 
 - (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust {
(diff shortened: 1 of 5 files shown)


kcharwood:
Thanks for putting this together! I'm still digging into it, but I'd like to see more negative test cases around `setValidatesCertificateChain` to `YES` and `NO`. We literally can't have *enough* tests in this area. The more the merrier.

If there is a good way to test those static functions you added, they would be good test candidates as well.

I'll keep digging in for more feedback.

AFNetworking/AFSecurityPolicy.m
```diff
@@ -294,6 +415,39 @@ - (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
 
             return trustedPublicKeyCount > 0 && ((self.validatesCertificateChain && trustedPublicKeyCount == [serverCertificates count]) || (!self.validatesCertificateChain && trustedPublicKeyCount >= 1));
         }
+        case AFSSLPinningModePublicKeyHash: {
+            NSUInteger trustedPublicKeyHashCount = 0;
+            NSArray *publicKeys = AFPublicKeyTrustChainForServerTrust(serverTrust); // Get Remote Public Keys
+
```
kcharwood: I think we need something like line 404 here where if aren't validating the the entire certificate chain, we are only looking at the leaf.

We could also use some tests around that. For instance if we only have an intermediate cert key, and not the leaf, and we aren't validating the cert chain, I think it should fail. With the current code, I think that test would pass. Thoughts?

kcharwood: Sample test:

```objective-c
- (void)testLeafPublicKeyHashPinningFailsForHTTPBinOrgPinnedCertificateAgainstHTTPBinOrgServerTrustChainWhenOnlyOneIntermediateKeyIsPresentAndNoValidatingCertificateChain {
    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKeyHash];
    
    [policy setPinnedPublicKeyHashes:[NSArray arrayWithObjects:@"SXxoaOSEzPC6BgGmxAt/EAcsajw=",
                                      nil]];
    [policy setValidatesCertificateChain:NO];
    
    SecTrustRef trust = AFUTHTTPBinOrgServerTrust();
    XCTAssertFalse([policy evaluateServerTrust:trust forDomain:nil], @"HTTPBin.org Public Key Hash Pinning Mode Should Have Failed");
    CFRelease(trust);
}
```

That test fails with the current code. Adding a check like line 404 in the public key hash section and then the test passes.

kcharwood:
A couple more tests I thought of:

```objective-c
- (void)testLeafPublicKeyHashPinningFailsForHTTPBinOrgPinnedCertificateAgainstHTTPBinOrgServerTrustChainWhenOneKeyHasOneWrongKey {
    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKeyHash];
    
    [policy setPinnedPublicKeyHashes:[NSArray arrayWithObjects:@"G2bwK+Vdcc6RPUi2h2A9mOPezwc=",
                                      @"SXxoaOSEzPC6BgGmxAt/EAcsajw=",
                                      @"CAEUNlja29wg0tQeD45uzUPUsiA=",
                                      @"wrongkey",
                                      nil]];
    [policy setValidatesCertificateChain:YES];
    
    SecTrustRef trust = AFUTHTTPBinOrgServerTrust();
    XCTAssertFalse([policy evaluateServerTrust:trust forDomain:nil], @"HTTPBin.org Public Key Hash Pinning Mode Should Have Failed");
    CFRelease(trust);
}

- (void)testLeafPublicKeyHashPinningFailsForHTTPBinOrgPinnedCertificateAgainstHTTPBinOrgServerTrustChainWhenOneKeyHasTooManyKeys {
    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKeyHash];
    
    [policy setPinnedPublicKeyHashes:[NSArray arrayWithObjects:@"G2bwK+Vdcc6RPUi2h2A9mOPezwc=",
                                      @"SXxoaOSEzPC6BgGmxAt/EAcsajw=",
                                      @"CAEUNlja29wg0tQeD45uzUPUsiA=",
                                      @"T5x9IXmcrQ7YuQxXnxoCmeeQ84c=",
                                      @"ExtraKey",
                                      nil]];
    [policy setValidatesCertificateChain:YES];
    
    SecTrustRef trust = AFUTHTTPBinOrgServerTrust();
    XCTAssertFalse([policy evaluateServerTrust:trust forDomain:nil], @"HTTPBin.org Public Key Hash Pinning Mode Should Have Failed");
    CFRelease(trust);
}
```

arein:
Thanks for taking the time to review!

I've added further tests as discussed.

cnoon:
Hi @arein, this looks like a great start. Nice work!

I have a few been playing catchup here with SPKI in general and have been digging through the [RFC 7469](http://tools.ietf.org/html/rfc7469) spec to get up to speed. From what I can tell, it looks as though the `sha1` assumption may be a bit too naïve. In [Section 2.4](http://tools.ietf.org/html/rfc7469#section-2.4), they point out that the pin consists of a known hashing algorithm which is specified. For example, it could be `sha1`, `sha256`, `md5`, etc. 

I don't believe this implementation accounts for anything other than `sha1` which seems problematic. If you are passing in the public key(s) to pin, it appears you would also need to specify the hashing algorithm used to generate those key(s). Then the implementation would have to account for all the possible (or likely) hashing algorithms. It would also have to be updated when a new hashing algorithm was released.

Could you possibly walk me through this a bit more? We want to make sure we nail this feature from all angles and use cases if we're going to roll it out.

arein:
Hi,
Thanks for pointing this out. The implementation is inspired by the implementation of Chromium (see [this](http://src.chromium.org/viewvc/chrome/trunk/src/net/http/transport_security_state_static.certs) and [this](http://src.chromium.org/viewvc/chrome/trunk/src/net/http/transport_security_state_static.json)) who also rely on sha1 hashes only.

I agree that I need to add the hash function to the hash. This would allow for additional hash algorithms in the future.

I've addressed this issue accordingly.

arein:
I've noticed a potential security threat in my implementation. Please let me review this before merging.

kcharwood:
Good notes @cnoon. I need to get more familiar with that RFC.

And @arein I'm in no rush to hit the merge button here. Security is the most important component of the entire framework to get right. I'll be soliciting lots of opinions to make sure we are good.

I also want to ramp up the negative test cases. It's more important that the security class FAILS as expected, rather than succeeds as expected. I want to really make sure we've nailed all the corner cases to make sure this are no security vulnerabilities.

cnoon:
I think I'm leaning towards something like the following for adding a public key as a pin:

```objc
// I'm not sure exactly which ones the spec currently supports, but I know it supports both sha1 and sha256
typedef NS_ENUM(NSUInteger, AFPublicKeyHashAlgorithm) {
    AFPublicKeyHashAlgorithmMD2,
    AFPublicKeyHashAlgorithmMD5,
    AFPublicKeyHashAlgorithmSHA1,
    AFPublicKeyHashAlgorithmSHA256
};

- (void)addPublicKey:(NSString *)publicKey generatedWithHashAlgorithm:(AFPublicKeyHashAlgorithm)hashAlgorithm;
```

Then you could switch on the hash algorithm in the implementation using the respective hashing implementation (`CC_SHA1`, `CC_SHA256`, `CC_MD5`, etc.).

Thoughts?

kcharwood:
@cnoon I'm wondering if we need to combine the concept of certificate chain validation and pinning into one class.

Right now you get different behavior depending on whether you have `validatesCertificateChain` set to NO or YES, and whether or not you have multiple keys/certs/key hashes pinned. I'm not a big fan of that design.

I'll take a stab at what it could be.

cnoon:
That sounds great @kcharwood. This solution should be a simple as possible to use and impossible to misuse.

kcharwood:
@cnoon there in lies the tricky part...

kcharwood:
This looks interesting:

https://github.com/WebKit/webkit/blob/06625bd2e6209996ddfa7acc44d3ce1e553b25da/Source/WebCore/platform/spi/cocoa/SecuritySPI.h#L49

```objective-c
#ifndef SecuritySPI_h
#define SecuritySPI_h

#if USE(APPLE_INTERNAL_SDK)
#include <Security/SecCertificatePriv.h>
#else

typedef uint32_t SecSignatureHashAlgorithm;
enum {
    kSecSignatureHashAlgorithmUnknown = 0,
    kSecSignatureHashAlgorithmMD2 = 1,
    kSecSignatureHashAlgorithmMD4 = 2,
    kSecSignatureHashAlgorithmMD5 = 3,
    kSecSignatureHashAlgorithmSHA1 = 4,
    kSecSignatureHashAlgorithmSHA224 = 5,
    kSecSignatureHashAlgorithmSHA256 = 6,
    kSecSignatureHashAlgorithmSHA384 = 7,
    kSecSignatureHashAlgorithmSHA512 = 8
};

#endif

#if (PLATFORM(MAC) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 101100) || (PLATFORM(IOS) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 90000)
EXTERN_C SecSignatureHashAlgorithm SecCertificateGetSignatureHashAlgorithm(SecCertificateRef certificate);
#endif

#endif // SecuritySPI_h
#endif
```
Found by Pete: https://twitter.com/steipete/status/601418212643856385

arein:
@kcharwood, @cnoon :

1. `validatesCertificateChain` only affects the pinning behaviour, right? Right now for `AFSSLPinningModePublicKey` when `validatesCertificateChain` is `YES` we expect the set of certificates retrieved from the server to __exactly__ match the predefined hashes or certificates. And if set to `NO`, it passes in case __any__ certificate matches. I think the naming is a bit confusing here. I would suggest renaming `validatesCertificateChain` to `requireAllPinnedCertificatesToBeMatched`. Then in every other case we should - from my point of view - change the logic so that __any__ pinned certificate passes. To my mind, we should leave the decision of whether the developer wants to pin the CA Cert, Intermediate, or leave certificate to herself.

2.  The current implementation of `AFSSLPinningModePublicKey` and `AFSSLPinningModePublicKeyHash` pin only the public key. This allows for a few options to attack (search for "SubjectPublicKeyInfo" in [this](https://www.imperialviolet.org/2011/05/04/pinning.html) blog post). It is not trivial to retrieve the SPKI from a certificate using Apple's Security Library. In fact, we need a DER decoding routine. I've been in touch with [Andrew Snare](https://github.com/asnare) who [once wrote a minimal DER encoder for this purpose](https://github.com/xebia/ssl-pinning-ios-example/blob/master/SSLTweets/DER/XBDERDecoder.m). [Andrew](https://github.com/asnare) is in touch with his previous employer on whether the decoder can be open-sourced. We were wondering exactly what kind of relicensing would be necessary to make this happen. Would you know about that?

3. I disagree about adding support for multiple hash algorithms at this point. This does not increase security. Would it be okay to add this at a later stage when/if required?

kcharwood:
Hey @arein 

On number 1, that's almost right. When `validatesCertificateChain` is set to NO and you are using `AFSSLPinningModePublicKey`, it is only matching the leaf public key. If you look at line [284](https://github.com/AFNetworking/AFNetworking/blob/2.5.4/AFNetworking/AFSecurityPolicy.m#L284), you'll see we are resetting the array to only an array with the first object, which in this case turns out to be the leaf. All that being said, I do think this code needs to be revisited. Writing up an in depth security guide has exposed a few issues for me that I think can be simplified and implemented in a more robust manner.
2. I don't know much on that front. That is getting into areas a bit murky for me. If you look at my post above, I'm wondering if Apple may have some new features for us in iOS 9 that may make this easier. If so, this could be a feature we add for iOS 9+. Obviously we'll know in 2 weeks :) I'm hoping those won't be private...
3. I don't have a firm opinion either way on this at the moment. I would definitely like some more input from the community.

arein:
Hey @kcharwood 

1. Following up on our discussion about the different behaviour of `validatesCertificateChain ` I found out that actually the current implementation of `AFSSLPinningModeCertificate` and `AFSSLPinningModePublicKeyHash` are semantically identical. I also noticed that we could change the implementation of `AFSSLPinningModePublicKey` to match `AFSSLPinningModeCertificate` and `AFSSLPinningModePublicKeyHash` while remaining backwards compatible simply by removing [line 284](https://github.com/AFNetworking/AFNetworking/blob/2.5.4/AFNetworking/AFSecurityPolicy.m#L284). Check this comparison on the effect of `validatesCertificateChain` on the various pinning modes:
  + `AFSSLPinningModeNone`: Not affecting this mode
  + `AFSSLPinningModeCertificate`:
    + If `validatesCertificateChain` is `NO`: __Any__ of the pinned certificates has to match (see [this discussion](https://developer.apple.com/library/mac/documentation/Security/Reference/certifkeytrustservices/index.html#//apple_ref/c/func/SecTrustEvaluate))

    + `validatesCertificateChain` is `YES`: __All__ pinned certificates have to match
  + `AFSSLPinningModePublicKey`: 
    + If `validatesCertificateChain` is `NO`: The public key of the __leaf__ server certificate has to match one of the pinned public keys

    + If `validatesCertificateChain` is `YES` all certificates have to match.

  + `AFSSLPinningModePublicKeyHash`: Semantics equal `AFSSLPinningModeCertificate`. 

2. Would be great to see improvements on this end.

In general, I noticed that actually setting `validatesCertificateChain` to `NO` already allows for backup keys to be specified. For that reason, I revoke my previous suggestion of renaming `validatesCertificateChain` to `requireAllPinnedCertificatesToBeMatched`.

kcharwood:
Thanks @arein 

Check out #2744. I am attempting to get community consensus on if we even need to worry about validating the whole chain. I haven't understood the security value one would get out of that.

In fact, the security engineer in that thread actually says to pin higher up the chain, and not even use the leaf. Definitely some interesting commentary there.

Whatever decision we come to in #2744 will definitely affect your implementation here.

arein:
Agreed. Let's see.

I'll keep updating the PR as I hear back from [Andrew](https://github.com/asnare).

asnare:
Sorry for taking so long to get back to you, but we're fine with relicensing the DER decoder that I wrote.

I should explain that I didn't license it as open-source originally because I was concerned about how robust/secure it was. In particular, I reasoned that it's code that would end up being an inviting attack surface for an adversary that was trying to compromise the certificate chain. I didn't think it was ethical to just put it _out there_ until I was reasonably confident it was safe to use. (DER-decoding is sufficiently esoteric that the _many-eyeballs-make-for-shallow-bugs_ argument doesn't necessarily hold.)

In the interim, this code has been used in several fairly high-profile applications that were subject to external security (code-level) reviews. Even though I wasn't personally involved in these projects, they give me sufficient confidence to assuage my concerns.

Please let me know if there's anything I need to do formally to effect the relicensing.

arein:
@asnare Thanks! [I've changed the copyright accordingly](https://github.com/arein/AFNetworking/commit/dc3d2513af42d4bb20fa2e7174d27586f13c9426#diff-6b42aee5819378d721a21ff296ef35d0). Would you happen to have tests for the Encoder?
@kcharwood Please review. Thanks to @asnare's encoder I am now confident with the implementation. I've also added a [gist](https://gist.github.com/arein/127c2756aa64c4b61c74) that shows how to create the public key hashes. The public key hashes are now equal to the public key hashes used in [OkHttp](http://square.github.io/okhttp/javadoc/com/squareup/okhttp/CertificatePinner.html). This allows for good interoperability.

arein:
@kcharwood Any take on this?

kcharwood:
Haven't forgotten about you! We actually spent some time talking this over with Apple Security engineers at WWDC and got some feedback. It's on my list to follow up with, just haven't gotten to it quite yet. Still hoping to get this in the 2.6 release, which should be some time in the next couple of weeks before we turn our focus to 3.0 and iOS 9.

asnare:
@arein I'm afraid I don't have any test-cases that I'm able to share. If these are desired, the basic strategy is:

1. Test a corpus of certificates. (Personally I'd use the contents of `/etc/ssl/certs` on any up-to-date Linux system as a convenient starting point.)
2. Fuzz the corpus, and ensure the code never crashes: it should always either yield the SPKI or a polite error code. (Fuzzing will not always corrupt a certificate sufficiently to prevent extraction of the SPKI information.) 
3. Manually adjust a good certificate to use the same key but a different algorithm identifier. (One of the key things that using the SPKI is intended to protect against is an attack where the key is reinterpreted as something different.)

(I'd love to do this myself but have limited time available at the moment.)

@kcharwood Hi Kevin, I'd love to hear the feedback from Apple's Security folk, if you're able to share it. (Privately, if necessary.)

arein:
@asnare Thanks. Will do!
@kcharwood Any updates on this?

kcharwood:
Ok just got the PR in that removed certificate chain validation. Now I can focus on this one...

kcharwood:
@arein would you want to rebase it off master?

kcharwood:
I've decided to push this one out until a future release. We have a lot of stuff ready to go with 2.6, so I'm in the final stages of prepping and pushing that out.

I think this still needs a bit more planning and thought before pulling it into master. Just want to make sure we're all 100% good with these changes, since we dont want to mess up security :)

arein:
I agree that we should not mess up security and I am fine with pushing it back as well. How should we proceed to make sure this just lays around forever? Who would be best to review this?

rist:
Hi
are there any updates regarding merging this in either 2.x or 3.x?

Are you confident enough about the security of this PR that you would recommend us merging it in our private branch? (We anyway have to do a penetration test of our app after updating SSL pinning and perhaps could share the pen test results)

revile:
Any updates on this? Has it been abandoned?

Pitometsu:
ping, @kcharwood, can it be merged? Any work/PR required for this?
 I need this feature in my current project, would be nice to have it in upstream here.

Pitometsu:
@asnare any updates?

Pitometsu:
https://www.owasp.org/index.php/Certificate_and_Public_Key_Pinning

Pitometsu:
Is it actual yet, or users should use separate libraries for that like https://github.com/datatheorem/TrustKit ?

asnare:
@Pitometsu I'm not really involved in the merge process. My involvement here is limited to the original example implementation, and specifically code to extract to the SPKI information from a certificate.

kcharwood:
Hey sorry for this to still be sitting here. I apologize for that.

I'm pretty conservative when it comes to changes/additions to the security policy flow. There are a lot of people who depend on this library to be secure, so changes should be heavily scrutinized.

I'm not comfortable merging something in that I don't _fully_ understanding here, which is why its still outstanding. I simply haven't had the time to get fully up to speed on this topic, and there hasn't been a ton of requests for this feature, deeming it low priority for me.

@cnoon Have you seen any requests for this on the AF side?

WirecardMobileServices:
@kcharwood Any chance of merging this feature into the master any time soon ? Public Key Hash pinning is available in other libraries as well. Having the PK hash does not expose the PK details or public certificate itself, plus having the backup PK hash in place does not expose the backup certificate in the case the app binary is decompiled. Would be awesome to have PK hash pinning in place

pmlbrito:
Is this feature still being considered? Its kind of messy to maintaining an external dependency to allow this type of pinning

jshier:
Closing due to age.

PR #2737, opened 2015-05-19
Allow same library for app and extensions.

bnickel:
clang will silence availability errors in classes and methods that are also unavailable to the target.  The following have been made unavailable to app extensions:
- `AFNetworkActivityIndicatorManager` class
- `UIAlertView` category methods
- `AFURLConnectionOperation setShouldExecuteAsBackgroundTaskWithExpirationHandler:`

Benefits of change:
- Fixes CocoaPods support for App Extensions.
- Allows single framework to be used in app.
- Eliminates the need for `AF_APP_EXTENSIONS`.

### DIFF
--- a/UIKit+AFNetworking/UIAlertView+AFNetworking.h
+++ b/UIKit+AFNetworking/UIAlertView+AFNetworking.h
@@ -23,7 +23,7 @@
 
 #import <Availability.h>
 
-#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && !defined(AF_APP_EXTENSIONS)
+#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
 
 #import <UIKit/UIKit.h>
 
@@ -48,7 +48,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 #if __IPHONE_OS_VERSION_MIN_REQUIRED >= 70000
 + (void)showAlertViewForTaskWithErrorOnCompletion:(NSURLSessionTask *)task
-                                         delegate:(nullable id)delegate;
+                                         delegate:(nullable id)delegate NS_EXTENSION_UNAVAILABLE_IOS("Not available in app extensions.");
 #endif
 
 /**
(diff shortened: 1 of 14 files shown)


bnickel:
For reference, this is meant to resolve the continuing CocoaPods 0.36+ issues mentioned in #2321.

kcharwood:
For #2749 as well

kcharwood:
Is it time we had an extension to the example project?

mattt:
We should go all out. App extension, watch app... The works.

jervine10:
Will this add support for native watchOS 2 apps as well?

bnickel:
WatchOS is a different story. I think it doesn't have UIKit so the iOS library from the project and the iOS cocoa pod would fail unless you just exclude the UIKit subspec. (Assuming cocoapods works in general.) These are just guesses though.

kcharwood:
Thanks for this @bnickel 

Would you want to add a Today Extension to the example project so this behavior can be verified going forward?

bnickel:
Sure

bnickel:
I've added a Today extension which validates that the code compiles with the annotated methods and classes.  I can't get the Travis failure to reproduce locally.

UIKit+AFNetworking/UIAlertView+AFNetworking.h
```diff
@@ -23,7 +23,7 @@
 
 #import <Availability.h>
 
-#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && !defined(AF_APP_EXTENSIONS)
+#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
```
kcharwood: Do you see any reason to not mark this whole class as `NS_EXTENSION_UNAVAILABLE_IOS`?

bnickel: I don't remember. I think I was just following the pattern of UIAlertView.h. If it compiles the other way, I'm sure it's fine.

kcharwood:
This is actually looks good. Just verified all the tests pass on my local machine.

Thanks for putting this together!

Issue #2739, opened 2015-05-20
Apps crash after on backgrounding after displaying any UIActivityIndicatorView


bnickel:
Commit 850e5e78503ee4a1170b1ac79139025ecf87a6a9 clobbers `-[UIActivityIndicatorView dealloc]` and `-[UIRefreshControl dealloc]`.  In addition to probably causing memory leaks, this eliminates the default behavior of unsubscribing from `UIApplicationDidEnterBackgroundNotification` so backgrounding the app causes a message to be sent to a zombie.

kcharwood:
eeekkk you are right. Let me look further into this.

bnickel:
I just added #2741 with a potential fix.

PR #2685, opened 2015-04-25
#2684 fix Crash on AFNetworkingOperationDidStartNotification 

hhfa008:
#2684 fix Crash on AFNetworkingOperationDidStartNotification

PR #2741, opened 2015-05-20
Fix issue #2739

bnickel:
This introduces associated objects so notifications can be cleared in dealloc without clobbering the class's dealloc.

### DIFF
--- a/Tests/Tests/AFUIActivityIndicatorViewTests.m
+++ b/Tests/Tests/AFUIActivityIndicatorViewTests.m
@@ -146,6 +146,7 @@
     [operation cancel];
 }
 
+// Tests issue #2739. -[UIActivityIndicatorView dealloc] removes an observer and clobbering it in a category creates a zombie reference.
 - (void)testBackgroundingDoesNotCauseCrashWithOperation {
     XCTestExpectation *expectation = [self expectationWithDescription:@"No Crash"];
     AFHTTPRequestOperation *operation = [self.operationManager
--- a/Tests/Tests/AFUIActivityIndicatorViewTests.m
+++ b/Tests/Tests/AFUIActivityIndicatorViewTests.m
@@ -146,4 +146,21 @@
     [operation cancel];
 }
 
+- (void)testBackgroundingDoesNotCauseCrashWithOperation {
+    XCTestExpectation *expectation = [self expectationWithDescription:@"No Crash"];
+    AFHTTPRequestOperation *operation = [self.operationManager
+                                         HTTPRequestOperationWithRequest:self.request
+                                         success:^(AFHTTPRequestOperation *operation, id responseObject) {
+                                         } failure:nil];
+    [self.activityIndicatorView setAnimatingWithStateOfOperation:operation];
+    self.activityIndicatorView = nil;
+    
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        [[NSNotificationCenter defaultCenter] postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
+        [expectation fulfill];
+    });
+    
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+}
+
 @end
--- a/UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
+++ b/UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
@@ -30,7 +30,7 @@
 #import "AFURLSessionManager.h"
 #endif
 
-@interface AFActivityIndicatorAnimator : NSObject
+@interface AFActivityIndicatorViewNotificationObserver : NSObject
 @property (readonly, nonatomic, weak) UIActivityIndicatorView *activityIndicatorView;
 - (instancetype)initWithActivityIndicatorView:(UIActivityIndicatorView *)activityIndicatorView;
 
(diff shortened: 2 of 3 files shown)


kcharwood:
Should we just swizzle out `dealloc` in the category? I'm always hesitant to swizzle unless there is a better option, but it seems like the cleaner route here.

bnickel:
Some quick googling of "swizzling dealloc" suggests it is not a good idea for production code.  Associated objects are a reliable way to get notified of an object's deallocation.

kcharwood:
I'll dig in a bit more. I can revert #2685 in the near term.

kcharwood:
First time I have it done this way, and it's actually not as bad as I thought it would be. I might want to change a few names, and can you add some test coverage to the refresh control as well? Also, drop a comment in those tests with this issue number so we can have a bread crumb trail pointing back here.

bnickel:
I tried testing the refresh control but I couldn't trigger a crash.  As far as I can tell it doesn't have any notifications so the only thing a test could prove is wrong would be hypothetical memory leaks.  We know clobbering `dealloc` is itself a problem so I don't think we need to expand that.

kcharwood:
Good info. Thanks

bnickel:
Added a comment with the issue number.

UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
```diff
@@ -38,14 +82,19 @@ - (void)setAnimatingWithStateOfTask:(NSURLSessionTask *)task {
...
     [notificationCenter removeObserver:self name:AFNetworkingTaskDidSuspendNotification object:nil];
     [notificationCenter removeObserver:self name:AFNetworkingTaskDidCompleteNotification object:nil];
-
+    
     if (task) {
         if (task.state != NSURLSessionTaskStateCompleted) {
+            
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wreceiver-is-weak"
+#pragma clang diagnostic ignored "-Warc-repeated-use-of-weak"
```
cnoon: How about pulling the `activityIndicatorView` into a local strong ref here to avoid the clang pragmas? It would be cleaner, safer and is more of a general practice. It shouldn't result in a retain cycle since it is only a [temporary retain](http://www.cocoawithlove.com/2009/07/rules-to-avoid-retain-cycles.html).

```objc
- (void)setAnimatingWithStateOfTask:(NSURLSessionTask *)task {
    UIActivityIndicatorView *activityIndicatorView = self.activityIndicatorView;
    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];

    [notificationCenter removeObserver:self name:AFNetworkingTaskDidResumeNotification object:nil];
    [notificationCenter removeObserver:self name:AFNetworkingTaskDidSuspendNotification object:nil];
    [notificationCenter removeObserver:self name:AFNetworkingTaskDidCompleteNotification object:nil];
    
    if (task) {
        if (task.state != NSURLSessionTaskStateCompleted) {
            if (task.state == NSURLSessionTaskStateRunning) {
                [activityIndicatorView startAnimating];
            } else {
                [activityIndicatorView stopAnimating];
            }

            [notificationCenter addObserver:self selector:@selector(af_startAnimating) name:AFNetworkingTaskDidResumeNotification object:task];
            [notificationCenter addObserver:self selector:@selector(af_stopAnimating) name:AFNetworkingTaskDidCompleteNotification object:task];
            [notificationCenter addObserver:self selector:@selector(af_stopAnimating) name:AFNetworkingTaskDidSuspendNotification object:task];
        }
    }
}
```

If we decide to go this route, it should be applied to all changes in this PR in both the `UIActivityIndicatorView` and the `UIRefreshControl`.

Thoughts?

Tests/Tests/AFUIActivityIndicatorViewTests.m
```diff
@@ -146,4 +146,22 @@ - (void)testOperationDidFinishNotificationDoesNotCauseCrashForAIVWithOperation {
...
+    [self.activityIndicatorView setAnimatingWithStateOfOperation:operation];
+    self.activityIndicatorView = nil;
+    
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1.0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        [[NSNotificationCenter defaultCenter] postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
+        [expectation fulfill];
+    });
+    
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+}
```
cnoon: This test definitely crashes if you clobber `dealloc`. Nice work!

cnoon:
Could you fix the `-(void)dealloc` spacing on both the animator classes? I know you didn't make the change, but we should fix it before I forget!

cnoon:
Overall this is a great set of changes from my POV. I only have the couple of things and I'm good with these changes @kcharwood. Thanks for this @bnickel!

kcharwood:
@bnickel I think I'd like to see the class names changed as well to better describe what they are actually doing.

I'm thinking `AFRefreshControlNotificationObserver` and `AFUIActivityIndicatorViewNotificationObserver`?

UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
bnickel: The reason I did it this way is simply that adding a strong variable adds no real value other than its indirect silencing of warnings and it seems like a pattern of this project to directly silence warnings on a one-off basis rather than adapt the code or build flags.

The warning "Weak property may be unpredictably set to nil" is still technically valid unless we do `if (activityIndicatorView)`.  This is unnecessary in our case because the only time it could not be there is in the async dispatches and sendmsg will handle `nil`.  `arc-repeated-use-of-weak` is really a compiler issue since its only ever used in once per branch.

Whatever y'all decide I'm fine with.

kcharwood:
Thanks for your work here @bnickel 

I'm going to merge this in now to squash this crash.

:beers:

PR #2745, opened 2015-05-22
Fixed UIActivityIndicatorView+AFNetworking.h crash bug.

E-B-Smith:
Fixed UIActivityIndicatorView+AFNetworking.h crash bug.

### DIFF
--- a/UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
+++ b/UIKit+AFNetworking/UIActivityIndicatorView+AFNetworking.m
@@ -95,15 +95,7 @@
 
 -(void)dealloc
 {
-    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 70000
-    [notificationCenter removeObserver:self name:AFNetworkingTaskDidCompleteNotification object:nil];
-    [notificationCenter removeObserver:self name:AFNetworkingTaskDidResumeNotification object:nil];
-    [notificationCenter removeObserver:self name:AFNetworkingTaskDidSuspendNotification object:nil];
-#endif
-    
-    [notificationCenter removeObserver:self name:AFNetworkingOperationDidStartNotification object:nil];
-    [notificationCenter removeObserver:self name:AFNetworkingOperationDidFinishNotification object:nil];
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
 @end


kcharwood:
Thanks for the PR. It's actually slightly more complicated than that.

Check out #2741 where we are tracking this fix. I welcome your feedback there.

E-B-Smith:
Why are you even discussing swizzling deallocs?  Can you give me a quick
TL:DR?

If this is an issue it seems like you may have some fundamental structural
issues in your code that should be addressed first.



On Fri, May 22, 2015 at 8:46 AM, Kevin Harwood <notifications@github.com>
wrote:
>
> Closed #2745.
>
> —
> Reply to this email directly or view it on GitHub.




--
===
::: E.B. Smith
::: smith.ed.b@gmail.com
::: gmail

kcharwood:
Thanks @E-B-Smith.

It's pretty complicated here. The notifications are being added to the UIKit objects from within a category, and the only "good" place to remove the observer is from `dealloc`. However, since we are in a category, we can't just override dealloc because that clobbers the actual implementation in UIKit, which is obviously bad.

Your fix here "works" because it actually removes the `UIApplicationDidEnterBackgroundNotification" observer as well, but any UIKit logic in dealloc is not called.

Swizzling is a decent approach to "hook" in from within a category, but dealloc is the only method that gets complicated with in ARC. My suggestion for swizzling dealloc probably wasnt the best approach there.

The current fix in #2741 goes for object association, where we CAN control dealloc, and safely remove the observers in our own class.

Sorry if that still doesnt make complete sense. I should have caught it in the original merge that we shouldnt be doing things in dealloc within a category.

E-B-Smith:
Sorry, i was too harassed and in a hurry this morning to be patient and
kind.

Do the object association.  That in fact I have a fix that I went with.  I
will submit it soon when I have time.

Why is this a category and not a subclass?






On Fri, May 22, 2015 at 9:59 AM, Kevin Harwood <notifications@github.com>
wrote:

> Thanks @E-B-Smith <https://github.com/E-B-Smith>.
>
> It's pretty complicated here. The notifications are being added to the
> UIKit objects from within a category, and the only "good" place to remove
> the observer is from dealloc. However, since we are in a category, we
> can't just override dealloc because that clobbers the actual implementation
> in UIKit, which is obviously bad.
>
> Your fix here "works" because it actually removes the
> `UIApplicationDidEnterBackgroundNotification" observer as well, but any
> UIKit logic in dealloc is not called.
>
> Swizzling is a decent approach to "hook" in from within a category, but
> dealloc is the only method that gets complicated with in ARC. My suggestion
> for swizzling dealloc probably wasnt the best approach there.
>
> The current fix in #2741
> <https://github.com/AFNetworking/AFNetworking/pull/2741> goes for object
> association, where we CAN control dealloc, and safely remove the observers
> in our own class.
>
> Sorry if that still doesnt make complete sense. I should have caught it in
> the original merge that we shouldnt be doing things in dealloc within a
> category.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/AFNetworking/AFNetworking/pull/2745#issuecomment-104713395>
> .
>



-- 
===
::: E.B. Smith
::: smith.ed.b@gmail.com
::: gmail

PR #2746, opened 2015-05-22
Changed Accept-Language header to start with the locale with country code

patcheng:
[NSLocale preferredLanguages] returns a list of 'primary-tag', e.g. en, fr, zh-Hans

That is not ideal when the server needs to know the country code of those locale.

So, prepend the header with a more explicit locale constructed from [NSLocale currentLocale]


kcharwood:
Locale code. It's amazing how convoluted this is on iOS. And neither of these approaches show the actual language the app is being presented in! You have to go to NSBundle mainBundle for that.

This is certainly important to get right, but I want to make sure we are getting it right here.

kcharwood:
Would you be able to add some tests here?

patcheng:
I re-did the change in https://github.com/AFNetworking/AFNetworking/issues/2888 (hopefully it's easier to follow)

kcharwood:
closing this for #2888

PR #2747, opened 2015-05-23
A better fix for the UIActivityIndicatorView+AFNetworking.m crash bug.

E-B-Smith:


E-B-Smith:
Might be better to subclass UIActivityIndicatorView but this fix works.

kcharwood:
Thanks @E-B-Smith 

I went with #2741 for now, since it handled both UIRefreshControl and UIActivityIndicatorView, and had tests to confirm the fix.

This fix is also interested, but I think we would need to extract that class into a private implementation somewhere so we could use it in both UIKit controls. I'd still be open to simplifying, but wanted to get #2741 in since it has been reviewed and confirmed, and many people have reported this issue.

:beers:

PR #2751, opened 2015-05-26
Add charset specification

yageek:
As `dataWithJSONObject:options:error:` returns an UTF-8 encoded
`NSData`, we can deduce the charset used for all requests


mattt:
According to [RFC 4627](http://tools.ietf.org/html/rfc4627), the `application/json` MIME type does not have a charset parameter. This is a common misconception, and was removed previously.

Issue #2744, opened 2015-05-22
SSL Pinning: What Should Be Certificate Chain Validation Expected Behavior?


kcharwood:
I've been spending some time deep in the `AFSecurityPolicy` class lately, and I'm wanting to get some feedback from the AFNetworking community on what the various needs are out there.

In AFNetworking, you can currently either pin against the leaf certificate, or the entire chain itself. There is no way to pin against only an intermediate certificate (B in the diagram below), or a subset of the chain (A and B in the diagram below). It's my understanding that there are some people who would want to only pin against an intermediate certificate, or a subset of chain, to allow for production/development certs, and the exchange of new certificates, while limiting the scope of authenticated connections to only be from the intermediate chain.

* A (root)
   * B
      * C (leaf)

The big question: **Is this something worth supporting?** Are there any major security cons to this approach?

Follow up question: **Should you be able to provide multiple pinning chains to check against?** For example, in #2733 there is the concept of providing a backup key should one be needed in the future. Would it make sense to expose a backup chain, and allow the developer to provide multiple valid chains to check (essentially ORing them together?) That could potentially make dev server/production server management a bit easier as well.

kcharwood:
#2646 is one way to solve this.

reedloden:
@kcharwood, as requested, questions answered below.

**Is this something worth supporting?**

Definitely. It's not a good idea at all to pin to an EE (end-entity aka leaf) cert except in very limited circumstances. EE certs are meant to be short-lived (only a few years at max), which means old clients will just stop working once they expire. As such, it's way better to pin to a set of root CAs or intermediate CAs (root CAs are greatly preferred, unless the intermediate CA is specific to your company/organization).

My comment in https://github.com/AFNetworking/AFNetworking/issues/1852#issuecomment-36199389 explains some of the finer points, including linking to several blog posts from experts in this field about how pinning should work.

**Are there any major security cons to this approach?**

Obviously, the higher you pin, the more possible it is for another (possibly malicious) entity to also get a certificate that would be valid under your pinning configuration. However, pinning to just a few root CAs is far better than permitting all 200+ root CAs to be able to MITM your connection. It's a risk trade-off. The closer you go to the actual cert, the more likely it is that you'll need to update more frequently and potentially have broken clients.

**Should you be able to provide multiple pinning chains to check against?**

[...]

PR #2756, opened 2015-05-29
Eliminated duplicated pinned certificates

kcharwood:
We have identified an issue that prevents `validatesCertificateChain` from validating the entire chain if a developer mistakenly pins duplicate certificates.

The fix is to simply ensure pinned certificates does not contain any duplicates. This does not represent an actual security vulnerability, as the duplicate certificates still need to be present in the certificate chain of the server trust.

In addition, this problem may also go away in a future version, if we deprecate the `validatesCertificateChain` property as discussed in #2744 

For now, we can put this fix into 2.5.5.

### DIFF
--- a/Tests/Tests/AFSecurityPolicyTests.m
+++ b/Tests/Tests/AFSecurityPolicyTests.m
@@ -238,7 +238,7 @@ static SecTrustRef AFUTTrustWithCertificate(SecCertificateRef certificate) {
 }
 
 - (void)testSettingDuplicateCertificatesProperlyRemovesTheDuplicate {
-    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKey];
+    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeCertificate];
     
     SecCertificateRef httpBinCertificate = AFUTHTTPBinOrgCertificate();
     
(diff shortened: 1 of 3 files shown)


Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -237,6 +237,39 @@ - (void)testPublicKeyPinningFailsForHTTPBinOrgIfNoCertificateIsPinned {
     CFRelease(trust);
 }
 
+- (void)testSettingDuplicateCertificatesProperlyRemovesTheDuplicate {
+    AFSecurityPolicy *policy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModePublicKey];
```
cnoon: Should this be `AFSSLPinningModeCertificate` since you are testing certs here rather than public keys? I realize it really doesn't matter under the hood since the type doesn't affect the removal of duplicate certs.

Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -333,6 +366,25 @@ - (void)testCertificatePinningFailsForHTTPBinOrgIfDomainNameDoesntMatch {
...
+                                    (__bridge_transfer NSData *)SecCertificateCopyData(httpBinCertificate),
+                                    (__bridge_transfer NSData *)SecCertificateCopyData(httpBinCertificate),
+                                    (__bridge_transfer NSData *)SecCertificateCopyData(httpBinCertificate)]];
+    
+    CFRelease(httpBinCertificate);
+    
+    [policy setValidatesCertificateChain:YES];
+    
+    SecTrustRef trust = AFUTHTTPBinOrgServerTrust();
+    XCTAssertFalse([policy evaluateServerTrust:trust forDomain:nil], @"HTTPBin.org Public Key Pinning Mode Failed");
```
cnoon: Probably want the failure comment to be directed towards `Certificate Pinning` instead.

kcharwood:
Added your feedback @cnoon :beers:

cnoon:
Looks good good to me @kcharwood...I'd say merge it!

sunuslee:
hi, kcharwood. great job:)

PR #2760, opened 2015-06-01
remove code never used

swtlovewtt:
no need this codes,program never go to there


kcharwood:
Hi @swtlovewtt 

This code can in fact be run. If you look [here](https://github.com/AFNetworking/AFNetworking/blob/2.5.4/AFNetworking/AFHTTPRequestOperation.m#L87), you can see that calling the response object for the first time causes the response to be parsed. If a parsing error occurs, `self.error` can be set there, triggering the condition to be true that you have removed.

:beers:

swtlovewtt:
soga, I know

PR #2771, opened 2015-06-05
Provide URL request to the success block when images are retrieved from cache

chrisgibbs:
When UIImageView+AFNetworking successfully retrieves an image from the cache, the URL request was not being provided to the success block along with the image. This meant that there was no way to compare retrieved image's URL with the URL that should be displayed.

### DIFF
--- a/Tests/Tests/AFUIImageViewTests.m
+++ b/Tests/Tests/AFUIImageViewTests.m
@@ -65,6 +65,7 @@
      placeholderImage:nil
      success:^(NSURLRequest *request, NSHTTPURLResponse *response, UIImage *image) {
          XCTAssertEqual(request, weakSelf.cachedImageRequest, @"URL requests do not match");
+         XCTAssertNil(response, @"Response should be nil when image is returned from cache");
          XCTAssertEqual(image, weakSelf.cachedImage, @"Cached images do not match");
          [expectation fulfill];
      }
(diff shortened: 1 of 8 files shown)


kcharwood:
Thanks @chrisgibbs. I think this makes sense. I'd like to see a test added to verify the behavior going forward.

Tests/Tests/AFUIImageViewTests.m
```diff
@@ -0,0 +1,75 @@
...
+}
+
+- (void)testSetImageWithURLRequestUsesCachedImage {
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Image view uses cached image"];
+    typeof(self) __weak weakSelf = self;
+    [self.imageView
+     setImageWithURLRequest:self.cachedImageRequest
+     placeholderImage:nil
+     success:^(NSURLRequest *request, NSHTTPURLResponse *response, UIImage *image) {
+         XCTAssertEqual(request, weakSelf.cachedImageRequest, @"URL requests do not match");
```
kcharwood: Should we assert that the response IS nil? I believe that is what indicates the image returned from cache, right?

chrisgibbs: Sounds reasonable. I'll make sure that's documented as well.

chrisgibbs: Should be good now.

kcharwood:
All tests pass locally. Thanks for the PR!

:beers:

Note I'll want to make sure this behavior change is explicitly called out in the Release Notes.

PR #2789, opened 2015-06-16
Cleanup xcode7 warnings

cutz:
There are a couple new warnings being produced by xcode7.  This pull request cleans up warnings around nil being passed into non-nullable parameters, as well as designated initializers that are not being overridden in subclasses.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -114,12 +114,7 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
     static AFNetworkReachabilityManager *_sharedManager = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        struct sockaddr_in address;
-        bzero(&address, sizeof(address));
-        address.sin_len = sizeof(address);
-        address.sin_family = AF_INET;
-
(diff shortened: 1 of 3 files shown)


AFNetworking/AFNetworkReachabilityManager.m
```diff
@@ -143,6 +138,25 @@ + (instancetype)managerForAddress:(const void *)address {
     return manager;
 }
 
+- (instancetype)init
```
benzguo: should this just be unavailable? an alternative:

```objc
- (instancetype)init NS_UNAVAILABLE
{
    return nil;
}
```

cutz: That is an option, albeit I hadn't realized the NS_UNAVAILABLE macro was available for this.  However, there is something sensible that the -init method can do, and that is setup reachability with the default address (exactly what the  singleton was/is doing).  I'm with fine with either approach for fixing this warning.

cutz:
The nullability warnings were ignored in commit 0bb9e9c49d24e78616968b138daccbb3f7478095.  If we end up wanting this -init change over the one in #2822 I'm happy to rebase just that change.  The remaining changes in this pull request have been covered elsewhere.

kcharwood:
I've decided to take #2822. We currently don't let people use `init`, use I feel like thats a bit cleaner.

Thanks!

:beers:

PR #2794, opened 2015-06-18
fix memory leak in NSURLSession handling

olegnaumenko:


AFNetworking/AFURLSessionManager.m
```diff
@@ -316,7 +316,8 @@ + (void)load {
             7) If the current class implementation of `resume` is not equal to the super class implementation of `resume` AND the current implementation of `resume` is not equal to the original implementation of `af_resume`, THEN swizzle the methods
             8) Set the current class to the super class, and repeat steps 3-8
          */
-        NSURLSessionDataTask *localDataTask = [[NSURLSession sessionWithConfiguration:nil] dataTaskWithURL:nil];
+        NSURLSession * session = [NSURLSession sessionWithConfiguration:nil];
```
alistra: `+ (NSURLSession * nonnull)sessionWithConfiguration:(NSURLSessionConfiguration * nonnull)configuration`

This method actually requires an argument according to iOS 9 sdk nullability

AFNetworking/AFURLSessionManager.m
```diff
@@ -316,7 +316,8 @@ + (void)load {
...
+        NSURLSessionDataTask *localDataTask = [session dataTaskWithURL:nil];
```
alistra: `- (NSURLSessionDataTask * nullable)dataTaskWithURL:(NSURL * nonnull)url`

this too requires a parameter

phoney:
This is a real bug that should be fixed. Apple's docs say

"The session object keeps a strong reference to the delegate until your app explicitly invalidates the session. If you do not invalidate the session, your app leaks memory."

alistra's comments are correct but the fix is trivial. Just add the default configuration when creating the session. I believe that @kcharwood wrote this code so maybe you want to fix this if @olegnaumenko isn't responding.

kcharwood:
I'll take a look at this soon. Thanks @phoney

kcharwood:
Rebased with some additional changes in 09f6cc1

kcharwood:
Pushed to master in 09f6cc1

Pushed to 3.0.0 with b2c449c

Thanks everyone! :beers:

Issue #2572, opened 2015-02-25
Crash on AFImageWithDataAtScale when loading image


matthewvelie:
I'm using "UIImageView+AFNetworking.h" to remotely load in some images into a UITableViewCell.  Randomly (about every 1/3 launches) the app will crash.  It'll either be a SIGABRT or an EXC_BAD_ACCESS in AFURLResponseSerialization in the method below.
```
static UIImage * AFImageWithDataAtScale(NSData *data, CGFloat scale) {
    UIImage *image = [[UIImage alloc] initWithData:data];

    return [[UIImage alloc] initWithCGImage:[image CGImage] scale:scale orientation:image.imageOrientation];
}
```
The crash usually outputs the error: `malloc: *** error for object 0x7fe662909050: double free
*** set a breakpoint in malloc_error_break to debug`

I've had this happen in my project randomly and been able to reproduce it in a small sample project, which I can supply if needed.

In the sample project I'm just doing this inside tableview:CellForRowAtIndexPath:
`[cell.cellImageView setImageWithURL:[NSURL URLWithString:@"http://www.nmnathletics.com//pics/160/107.gif"]];`

I'm unsure if this is related to: https://github.com/AFNetworking/AFNetworking/issues/2334, I noticed someone on that mentioning that it happened with images for them, but for everyone else it seemed to be a different issue.

So far I've only been able to make this happen in the simulator, on a retina device.

mattt:
> I've had this happen in my project randomly and been able to reproduce it in a small sample project, which I can supply if needed.

Yes, please share that if you can.

matthewvelie:
Here's a quick sample project. https://www.dropbox.com/s/zb7n7fcw20k4mr1/testApp.zip?dl=0
It worked the first two times I ran it in the simulator, third time it ran, crashed soon as it tried to display the images.

tewha:
Crashed for me, too, but only once. I wish I could get it to happen again…

tewha:
It took maybe 50 runs before it crashed again. Here's what I found: The data is correct, even on this failed run, so it isn't a question of the server returning something invalid. The crash is on the first line of AFImageWithDataAtScale, [[UIImage alloc] initWithData:data]: EXC_BAD_ACCESS EXC_I386_GPFLT. I'm going to bang out a quick app to see if the problem is reproducible without AFNetworking.

tewha:
[...]

PR #2815, opened 2015-06-26
Fix for issue #2572 Crash on AFImageWithDataAtScale when loading image

paulosotu:
#2572 - UIImage initWithData is not thread safe. In order to avoid
crash we need to serialise call.

### DIFF
--- a/AFNetworking.xcworkspace/contents.xcworkspacedata
+++ b/AFNetworking.xcworkspace/contents.xcworkspacedata
@@ -136,6 +136,12 @@
       <FileRef
          location = "group:UIRefreshControl+AFNetworking.m">
       </FileRef>
+      <FileRef
+         location = "group:/Users/pferreira/work/GitHub/AFNetworking/UIKit+AFNetworking/UIImage+AFNetworking.h">
+      </FileRef>
+      <FileRef
+         location = "group:UIImage+AFNetworking.m">
+      </FileRef>
    </Group>
    <FileRef
       location = "group:Tests/AFNetworking Tests.xcodeproj">
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -530,18 +530,26 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
 #import <CoreGraphics/CoreGraphics.h>
 
 NSLock* _AFImageLock = nil;
-@interface ImageLockInitializer : NSObject
+@interface ImageSyncronizer : NSObject
 @end
-@implementation ImageLockInitializer : NSObject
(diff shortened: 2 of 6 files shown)


mfclarke:
I ran into this issue as well and tried using the changes from your pull request to fix. Please note that your +[ImageLockInitializer initialize] method never gets called. You need to use the class somewhere for it's initialize method to be called.

paulosotu:
You are right. Just committed a fix for it. Renamed  ImageLockInitializer to ImageSyncronizer and used its lock and unlock static method, this should do the trick.

chadpod:
Running into this one pretty frequently. Is this something that will be merged and released in the near future? Thanks.

kcharwood:
Anyone have any ideas how we could create a test that causes this to crash in the current version of the library?

paulosotu:
@kcharwood Probably iterating for example 10000 times to create 10000 threads all calling the method AFImageWithDataAtScale may do the trick... although being a threading issue it might sometimes not crash...

kcharwood:
I'm experimenting with a test like this right now:

```
- (void)testCrash {
    // This is an example of a functional test case.
    NSString *bundlePath = [[NSBundle bundleForClass:[AFImageResponseSerializerTests class]] resourcePath];
    NSString *imagePath = [bundlePath stringByAppendingPathComponent:@"huge_image.jpg"];
    NSData *data = [NSData dataWithContentsOfFile:imagePath];
    NSDictionary *headers = @{@"Content-Length":@"8621961",@"Content-Type":@"image/jpeg"};
    NSHTTPURLResponse *response = [[NSHTTPURLResponse alloc] initWithURL:[NSURL URLWithString:@"http://test.com"] statusCode:200 HTTPVersion:@"1.1" headerFields:headers];
    
    for (int i = 0; i < 100000; i++) {
        XCTestExpectation *expectation = [self expectationWithDescription:@"wait"];
        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
            AFImageResponseSerializer *serializer = [AFImageResponseSerializer serializer];
            XCTAssert([serializer responseObjectForResponse:response data:data error:nil]);
            [expectation fulfill];
        });
    }
    [self waitForExpectationsWithTimeout:30.0 handler:nil];
}
```

kcharwood:
@paulosotu Thanks for tackling this.

I made some minor changes [here](https://github.com/AFNetworking/AFNetworking/commit/4cf23fa1543fa1fa5b093b0b6d6d7c1513d7a419). If these looks good, I'll get this merged in!

Let me know your thoughts.

:beers:

paulosotu:
@kcharwood Glad to help :)

Seems fine to me, so go for it.

kcharwood:
Merged as part of #2860 

:beers: all around.

PR #2822, opened 2015-07-02
Fix "method override not found" warnings on Xcode 7b2

benzguo:
Fixes the `Method override for the designated initializer of the superclass '-init' not found` warnings on Xcode 7 beta 2. 

This is an alternative to https://github.com/AFNetworking/AFNetworking/pull/2789 , if we think `init` should be unavailable in these cases.


kcharwood:
Thanks for the PR @benzguo 

I've decided to go with #2789 since currently people should not be using the `init` initializers for those classes.

:beers:

kcharwood:
Typo on my part! I've decided to go with this change instead of #2789. 

:beers:

PR #2831, opened 2015-07-07
Add startMonitoring call to README under block based reachability call

jawwad:
Add startMonitoring call to README to clarify that startMonitoring still need to be called when using the block based reachability callback method


kcharwood:
:beers:

PR #2836, opened 2015-07-09
Updating AFNetworkActivityManagerTests to use a NiceMock

kcharwood:
We have lots of random race condition failures that I believe are caused by a mock class getting a notification from another request finishing from a previous test. I think if we switch over to a niceMock, that should squash that random failure.


kcharwood:
Didn't work.

PR #2845, opened 2015-07-17
Removed NSNull objects from JSON arrays

harlanhaskins:
Right now, `{"key": [1,2,null,4]}` becomes `@{@"key": @[@1, @2, [NSNull null], @4]}`.

This PR removes NSNulls from arrays while processing arrays in `AFJSONObjectByRemovingKeysWithNullValues`.

This PR also has a test for that behavior.

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -60,7 +60,9 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
     if ([JSONObject isKindOfClass:[NSArray class]]) {
         NSMutableArray *mutableArray = [NSMutableArray arrayWithCapacity:[(NSArray *)JSONObject count]];
         for (id value in (NSArray *)JSONObject) {
-            [mutableArray addObject:AFJSONObjectByRemovingKeysWithNullValues(value, readingOptions)];
+            if (![value isKindOfClass:[NSNull class]]) {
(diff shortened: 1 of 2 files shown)


harlanhaskins:
I think the `testDomainReachabilityNotification` test is failing on master.

kcharwood:
I'm going back and forth on what should be expected here. Should we strip the null in that case?

cc @kylef @cnoon

kylef:
I'm against this, a null value could result in the index of an item in an array to be moved, the position of an item within an array may be important and have semantic meaning in a given domain.

Take for example, in the specification for [Refract](https://github.com/refractproject/refract-spec) (a recursive data structure for expressing complex structures) there is a [compact structure](https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#compact-element-array). The compact structure should always be 4 elements within an array, the first being the element name, second being metadata, third being attributes and the fourth being the content. If the meta data was empty (null) and AFNetworking silently removed the 2nd value, the third value (attributes) would now be in the second values place and silently become metadata and it's value would change meaning.

#### Full Object

```json
["number", {"title": "Age", "description": "The users age"}, {"default": 10}, 5]
```

#### Problematic Case

```json
["number", null, {"default": 10}, 5]
```

The above would silently be converted to `["number", {"default": 10}, 5]` where the meaning of the semantics is changed.

harlanhaskins:
That's very fair. I can see that. Though, in my opinion, using position in an array for meaning a poor design choice.

I'll close this PR.

PR #2856, opened 2015-07-24
Removed Certificate Chain Validation

kcharwood:
As discussed in #2744, I have removed certificate chain validation as a feature in `AFSecurityPolicy`. After a community discussion as well as a conversation with the security team at Apple, there are no specific advantages to maintaining this feature going forward.

Starting with this pull request, a security policy will allow a server trust if any of the pinned certificates meeting the security requirements of the policy. 

In addition, I have completely refactored the security policy tests, and grouped them in a way that is more readable and more maintainable going forward. I would love some eyes on those tests and see if we can come up with any more, especially negative test cases.

We'll get this in first, then move on to bringing in public key hash pinning.

### DIFF
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -196,7 +196,6 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
         return nil;
     }
 
-    self.validatesCertificateChain = YES;
     self.validatesDomainName = YES;
 
     return self;
@@ -204,7 +203,7 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
 
 - (void)setPinnedCertificates:(NSArray *)pinnedCertificates {
     _pinnedCertificates = [[NSOrderedSet orderedSetWithArray:pinnedCertificates] array];
-    
+
     if (self.pinnedCertificates) {
         NSMutableArray *mutablePinnedPublicKeys = [NSMutableArray arrayWithCapacity:[self.pinnedCertificates count]];
         for (NSData *certificate in self.pinnedCertificates) {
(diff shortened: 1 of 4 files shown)


Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -1,29 +1,16 @@
...
+//  Created by Kevin Harwood on 7/24/15.
+//  Copyright (c) 2015 AFNetworking. All rights reserved.
 //
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
```
cnoon: We probably don't want to change the copyright header do we?

Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -109,520 +96,435 @@ static SecCertificateRef AFUTSelfSignedCertificateWithDNSNameDomain() {
...
 static SecTrustRef AFUTTrustWithCertificate(SecCertificateRef certificate) {
     NSArray *certs  = @[(__bridge id)(certificate)];
-
+    
     SecPolicyRef policy = SecPolicyCreateBasicX509();
     SecTrustRef trust = NULL;
     SecTrustCreateWithCertificates((__bridge CFTypeRef)(certs), policy, &trust);
     CFRelease(policy);
-
+
```
cnoon: I believe all the AFN source code has whitespace stripped. Do you want to continue to do that to avoid the tabbed diffs?

AFNetworking/AFSecurityPolicy.m
```diff
@@ -229,6 +228,19 @@ - (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust {
 - (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
                   forDomain:(NSString *)domain
 {
+    if (domain && self.allowInvalidCertificates && (self.SSLPinningMode == AFSSLPinningModeNone || [self.pinnedCertificates count] == 0)) {
```
cnoon: Is the `allowInvalidCertificates` supposed to be `validatesDomainName` in this change? It seems to me that allowing invalid certs and not using pinning should evaluate to `YES`.

Tests/Tests/AFSecurityPolicyTests.m
kcharwood: Good catch. By product of creating a new file for the new tests, then blasting the old one upon completion. I'll add it back.

kcharwood:
Going once.... Going Twice...

urklc:
Hi, when allowing invalid certificates, why do we need to pin the certificate only if **validatesDomainName** is true? There's no mention about domain name validation in Apple Docs referenced in this change?

PR #2761, opened 2015-07-24
problem with baseURL

kcharwood:
```Objective-C
/**
 The URL used to monitor reachability, and construct requests from relative paths in methods like `requestWithMethod:URLString:parameters:`, and the `GET` / `POST` / et al. convenience methods.
 */
@property (readonly, nonatomic, strong) NSURL *baseURL;
```

In the `- (instancetype)initWithBaseURL:(NSURL *)` method, `self.reachabilityManager` doesn't use the baseURL. I know I can adjust `self.reachabilityManager` later with `"initWithReachability`. But the description confused me. Is this a problem?


kcharwood:
Thanks!

:beers:

PR #2860, opened 2015-07-27
Fixing AFImageWithDataAtScale race condition crash

kcharwood:
As discussed in #2815 and #2572, there is a crash related to a race condition with `imageWithData:`  This PR takes in #2815 with a few minor modifications to hopefully squash this one.


egold:
Thanks for the fix. Would you consider releasing a 2.5.5 release w/ this fix instead of waiting for 2.6.0?

alandeguz:
I'm with @egold ; a 2.5.5 release would be most welcome.

tsabend:
:+1: for patch release

egold:
2.6.0 is out, with the fix!

rnewman:
Has anyone filed a rdar for this behavior of `UIImage`? The workaround we're using in Firefox for iOS is to create an intermediate `CIImage`, rather than to serialize, but it'd be nice to kill the workaround one day.

UIKit+AFNetworking/UIImage+AFNetworking.h
```diff
@@ -0,0 +1,35 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)
+
+#import <UIKit/UIKit.h>
+
+@interface UIImage (AFNetworking)
+
++ (UIImage*) safeImageWithData:(NSData*)data;
```
pronebird: is it really without a prefix here? I might have overlooked something but it seems that it is defined as `af_safeImageWithData` above.

0xced: The method name in the header does not match the one in the implementation (`af_safeImageWithData`).

kcharwood: Ya we'll need to correct that, and probably force everything through the main thread.

benasher44: @kcharwood has anyone experimented with CGImageSource as a way of getting from data to a CGImage? I can't find any documentation that explicitly says this way would be thread-safe, but then again I haven't found any that says +[UIImage imageWithData:] is either.

benasher44: @kcharwood I was able spend some time comparing a method like this:

```
+ (UIImage *)imageFromImageSourceWithData:(NSData *)data {
  CGImageSourceRef imageSource = CGImageSourceCreateWithData((__bridge CFDataRef)data, NULL);
  CGImageRef imageRef = CGImageSourceCreateImageAtIndex(imageSource, 0, NULL);
  CFRelease(imageSource);
  UIImage *image = [UIImage imageWithCGImage:imageRef];
  CGImageRelease(imageRef);
  return image;
}
```

to just plain old ```[UIImage imageWithData:]``` So far, results show the above method is either as fast, faster, or sometimes slower than ```[UIImage imageWithData:]```. Basically, they're roughly equivalent performance-wise, depending on how you run the test. I'm going to play with it some more on a real device next week. But in the mean time, do you have a way to reliably make ```[UIImage imageWithData:]``` crash? Even just a some example environment settings (iPhone 5, iOS 8.4, etc.) would be helpful. Thanks in advance!

cnoon: I tried for quite some time in AlamofireImage to produce a crash for this scenario and was unsuccessful. There are several reports in the AFN issues that "claimed" they could make it crash in a test suite by iterating 1000's of times over image loading, but I couldn't reproduce the crash behavior.

kcharwood: @benasher44 

This thread has multiple people taking a [stab at it](https://github.com/AFNetworking/AFNetworking/issues/2572#issuecomment-76719795). There is a dropbox link that is no longer valid :(

benasher44: @kcharwood ah yes I see. Alright. I'll update if I find anything useful.

PR #2863, opened 2015-07-28
Add methods with completion blocks that complement setImageWithURL: &  setImageWithURL:placeholderImage:

chrsmys:
This is extremely useful for adding masks and filters to the image once it is retrieved.

### DIFF
--- a/UIKit+AFNetworking/UIImageView+AFNetworking.h
+++ b/UIKit+AFNetworking/UIImageView+AFNetworking.h
@@ -80,6 +80,21 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  Asynchronously downloads an image from the specified URL, and sets it once the request is finished. Any previous image request for the receiver will be cancelled.
+ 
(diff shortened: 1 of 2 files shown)


kcharwood:
Hi @chrsmys 

Thanks for the PR. Can you explain why the current method in the API doesn't fit your needs?

```Objective-C
- (void)setImageWithURLRequest:(NSURLRequest *)urlRequest
              placeholderImage:(nullable UIImage *)placeholderImage
                       success:(nullable void (^)(NSURLRequest *request, NSHTTPURLResponse *response, UIImage *image))success
                       failure:(nullable void (^)(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error))failure;
```

chrsmys:
It suits my needs, but this is for convenience. Creating a NSURLRequest seems unnecessary when there is already something that handles that for me.

kcharwood:
Gotcha. My gut is to not take it to prevent adding more methods to the API. It should be easy enough for you to wrap that method for convenience if you need it. For now I'm going to decline this, but if there is enough demand I'm happy to revisit it.

Thanks for the PR!

PR #2865, opened 2015-07-28
prevent NSJSONSerialization from adding extra backslashes to URL strings

scarlettche:
By add these code ,I can prevent NSJSONSerialization from adding extra backslashes to URL strings

### DIFF
--- a/AFNetworking/AFURLRequestSerialization.m
+++ b/AFNetworking/AFURLRequestSerialization.m
@@ -1265,7 +1265,11 @@ typedef enum {
             [mutableRequest setValue:@"application/json" forHTTPHeaderField:@"Content-Type"];
         }
 
-        [mutableRequest setHTTPBody:[NSJSONSerialization dataWithJSONObject:parameters options:self.writingOptions error:error]];
+        NSData *data = [NSJSONSerialization dataWithJSONObject:parameters options:self.writingOptions error:error];
+        NSString *dataStr = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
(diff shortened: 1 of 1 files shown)


kcharwood:
Can you please add tests demonstrating and verifying the solution?

Thanks!

kcharwood:
@scarlettche Would you want to add some tests to this PR? Thanks!

kcharwood:
Another bump @scarlettche. Can you show me the problem you are trying to solve. Thanks!

kcharwood:
I'm going to close this one out for now. If you would like to revisit, please do let me know here! Thanks!

:beers:

rodmaz:
@kcharwood This is the [problem](http://stackoverflow.com/questions/19651009/how-to-prevent-nsjsonserialization-from-adding-extra-escapes-in-url) this fix tries to solve.
iOS `NSJSONSerialization` escapes slashes and when used in HTTP body, this creates a whole lot of issues.

kylef:
This should not be a problem though, the JSON specification states that any character may be escaped, including a forward slash.

It's common that many JSON serialises escape forward slashes for compatibility reasons.

See [RFC 7159 Section 7](https://tools.ietf.org/html/rfc7159#section-7):

> Any character may be escaped.

rodmaz:
You are correct. I think it is an issue when (as in my case) a digest mechanism like MD5 is used in some sort of authentication mechanism. If the validation on server-side disregards the original JSON and the original escaping used (e.g. by parsing the JSON on server before the digest calculation), it may require some care on server-side.

PR #2871, opened 2015-08-01
Add "text/html" and "text/javascript" type to acceptableContentTypes

devSC:


kcharwood:
Adding non-standard acceptable content types is a slippery slope. Not all HTML or javascript is XML compatible, so dropping this in here potentially causes other issues. For your use, you can easily update the acceptable content types for each serializer, so there is no need to patch it here.

Thanks!

PR #2872, opened 2015-08-03
Added functionality to have send empty arrays in the params of a NSUR…

stephenyao:
When sending a request with params that contains an empty array of elements, the outgoing params would simply drop the field with empty array. This pull request fixes that by putting an empty array into the params of the outgoing request.


kcharwood:
Hey @stephenyao.

Would you be able to add some tests here?

stephenyao:
have opened another pull request

PR #2875, opened 2015-08-04
Added functionality to send empty array in params for NSURLRequest.

stephenyao:
Previously if trying to send a NSURLRequest with an empty array in the params the empty array would simply be dropped, now it will send the empty array as a key value pair with the value being an empty string.


stephenyao:
@kcharwood Hi Kevin, wondering if you've had a chance to take a look at this pull request?

stephenyao:
@kcharwood Any chance of merging this one in? Thanks

kcharwood:
I'm hesitant to merge this in, since it would be breaking change for current users of the library. Is there an easy this could be accomplished via subclassing?

kcharwood:
Dupe of #2763

PR #2878, opened 2015-08-04
remove unnecessary if statement

Jianing93:


CyberMew:
Won't the failure completion block not be called?

Jianing93:
The failure block will be called when `self.error != nil`. See line 120.

PR #2886, opened 2015-08-06
Version 2.5.4

rkim-gpfw:
Enhancements for UIImageView+AFNetworking:
    * Ability to set maximum concurrent operations for image network fetches
    * Ability to set operastion priority for a particular image network fetch
    * Determine if image is already cached


kcharwood:
Thanks for the PR! This is probably best managed by moving to a model that decouples the image downloading from the image view, similar to the architecture used in AlamofireImage.

I'll be migrating to something similar for 3.0 here as we deprecate NSURLConnection support.

PR #2894, opened 2015-08-12
watchOS_support: Fix to allow proper compilation on iOS SDK prior to 9.0

neildaniels:
The current `watchOS_support` branch does not correctly compile when trying to use an iOS SDK prior to 9.0. This issues is an annoying and avoidable regression.

The main issue is that the current code assumes that `TARGET_OS_IOS` is sufficient to determine that you’re compiling for iOS. However, this was only added in the 9.0 SDK, so it as absent when trying to use an older SDK (or Xcode version), which causes the compilation to take the “OS X” route.

The change in this PR uses the fact that `TARGET_OS_IPHONE` is true for both iOS and watchOS and exists prior the iOS 9 SDK.

### DIFF
--- a/AFNetworking/AFNetworking.h
+++ b/AFNetworking/AFNetworking.h
@@ -29,7 +29,7 @@
     #import "AFURLRequestSerialization.h"
     #import "AFURLResponseSerialization.h"
     #import "AFSecurityPolicy.h"
-#if !defined(TARGET_OS_WATCH) || !TARGET_OS_WATCH
+#if !TARGET_OS_WATCH
     #import "AFNetworkReachabilityManager.h"
     #import "AFURLConnectionOperation.h"
     #import "AFHTTPRequestOperation.h"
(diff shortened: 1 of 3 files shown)


neildaniels:
Technically, the `AFSecurityPolicy.m` changes could be simplified to use only `TARGET_OS_IPHONE` (since it’s true for both iOS and watchOS compilation), but the misleading name makes it seem like watchOS would be excluded.

kcharwood:
Hey Neil,

Thanks for the PR. If you look at the podspec I have in that current branch, you'll see we handle that with some [pch file additions](https://github.com/AFNetworking/AFNetworking/blob/watchOS_support/AFNetworking.podspec#L15). 

```Ruby
  pch_AF = <<-EOS
#ifndef TARGET_OS_IOS
  #define TARGET_OS_IOS TARGET_OS_IPHONE
#endif

#ifndef TARGET_OS_WATCH
  #define TARGET_OS_WATCH 0
#endif
EOS
  s.prefix_header_contents = pch_AF
```

Since Cocoapods is still the official distribution mechanism of AFNetworking, we should be good to go there. You can see tests are passing in that branch from 7.1+.

I'd rather go this route because it is more readable, rather than relying on `TARGET_OS_IPHONE`, which as you say is really misleading. If you choose to install the library in a different manner, you'll need to do some redefinition like I have done above.

:beers:

PR #2837, opened 2015-08-18
WatchKit support

kcharwood:
Including AFNetworking in a WatchKit 2 project results in a SystemConfiguration.h file not found error. 

I am seeing this in Xcode 7 / WatchOS 2 (beta 3)

### DIFF
--- a/Example/AFNetworking watchOS Extension/Assets.xcassets/README__ignoredByTemplate__
+++ b/Example/AFNetworking watchOS Extension/Assets.xcassets/README__ignoredByTemplate__
@@ -1,0 +1,1 @@
+Did you know that git does not support storing empty directories?
--- a/Example/AFNetworking watchOS Extension/InterfaceController.m
+++ b/Example/AFNetworking watchOS Extension/InterfaceController.m
@@ -1,0 +1,38 @@
+//
+//  InterfaceController.m
+//  AFNetworking watchOS Extension
+//
+//  Created by Kevin Harwood on 8/3/15.
+//  Copyright © 2015 Gowalla. All rights reserved.
+//
+
+#import "InterfaceController.h"
+
+
+@interface InterfaceController()
+
+@end
+
+
+@implementation InterfaceController
+
+- (void)awakeWithContext:(id)context {
+    [super awakeWithContext:context];
+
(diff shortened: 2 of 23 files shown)


Example/AFNetworking watchOS Extension/Assets.xcassets/README__ignoredByTemplate__
```diff
@@ -0,0 +1 @@
+Did you know that git does not support storing empty directories?
```
cnoon: Should this be committed? Where did this file even come from?

Example/AFNetworking watchOS Extension/InterfaceController.m
```diff
@@ -0,0 +1,38 @@
...
+}
+
+- (void)didDeactivate {
+    // This method is called when watch view controller is no longer visible
+    [super didDeactivate];
+}
+
+@end
+
+
```
cnoon: Newline cleanup at the top and bottom.

Example/AFNetworking watchOS Extension/Assets.xcassets/README__ignoredByTemplate__
kcharwood: I'm guessing xcode generated it when I added a watch target.

cnoon:
My only other thing @kcharwood would be whether the project has actually been updated with the new deployment targets in the podspec. It would be a bummer to throw a bunch of compiler warnings for everyone that picks up the new release if there are any new deprecation warnings. Have you had a chance to see if that causes any problems?

XenonChau:
```objective-c
- (NSURLRequest *)requestBySerializingRequest:(NSURLRequest *)request
                               withParameters:(id)parameters
                                        error:(NSError *__autoreleasing *)error
```

while "parameters" was converted into JSON or encrypted with AES.
the following assignment `query = AFQueryStringFromParametersWithEncoding(parameters, self.stringEncoding);` will cause an issue.

I don't know how many kinds of parameters. I'm using `[parameters isKindOfClass[NSDictionary class]] || [parameters isKindOfClass[NSArray class]] ` to avoid these two situations.

Is there any else?

PR #2908, opened 2015-08-19
Convert query escapes to `stringByAddingPercentEncodingWithAllowedCharacters`

kcharwood:
Since watchOS requires us to use `stringByAddingPercentEncodingWithAllowedCharacters`, we've migrated over to that. We're also using the logic that has been baked in to AlamoFire.

Note this does have one backwards compatible issue. `/` and `?` are no longer encoded in the query parameters, per RFC 3986.

```
 In RFC 3986 - Section 3.4, it states that the "?" and "/" characters should not be escaped to allow
 query strings to include a URL. Therefore, all "reserved" characters with the exception of "?" and "/"
 should be percent-escaped in the query string.
    - parameter string: The string to be percent-escaped.
    - returns: The percent-escaped string.
```

@cnoon want to take a look?

### DIFF
--- a/AFNetworking/AFURLRequestSerialization.m
+++ b/AFNetworking/AFURLRequestSerialization.m
@@ -61,26 +61,26 @@ static NSString * AFBase64EncodedStringFromString(NSString *string) {
     return [[NSString alloc] initWithData:mutableData encoding:NSASCIIStringEncoding];
 }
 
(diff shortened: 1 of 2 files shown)


cnoon:
This looks good to me dude! 👍🏼

onmyway133:
It would be better if we can have a flag for that :)

PR #2913, opened 2015-08-20
Remove unnecessary certificate validation

p4checo:
When full certificate chain validation was removed, the code to validate the complete chain wasn't removed but slightly modified so that only 1 certificate had to be present. This validation is redundant, because the pinned certificates are already set as anchors for the validation made in the previous 
steps.

In the case of certificates with multiple certification paths, this sometimes caused validations to fail which then required adding extra certificates to the `pinnedCertificates`.


kcharwood:
Thanks @p4checo. Could you add a test describing the failure above, and show it passing with the patch?

p4checo:
Hi @kcharwood, I've run tests in both `iOS Tests` and `OS X Tests` targets and they all pass locally, however `OSX Tests` fail in Travis CI and I don't understand why. Do you know why this may be happening?

p4checo:
Well, I'm having some difficulties in managing to create a `SecTrustRef` that replicates a server containing multiple validation chains. This is simply a chain that is "incomplete", and can be validated using one or more Root CA's.

I'm feeding the correct chain to `SecTrustCreateWithCertificates()` , but internally it completes the chain with one of the Root CA's (probably from the keychain), so only this Root CA's works when pinned.

p4checo:
Ok, I finally figured it out :relieved:. Turns out the problem wasn't related with the multiple validation paths, but with certificates with incomplete chains (i.e. no Root CA provided by the server).

In the process, I also figured out that iOS and OS X Security API's work in different ways, and sometimes return different results when validating the same certificate under the same conditions. Wish I didn't have to learn that the HARD way lol.

So, since the issue is in fact different, I'll close this PR and submit a new one. Sorry for any inconvenience. :smile:

Issue #2929, opened 2015-08-28
Dispatching success/failure blocks in "http_request_operation_processing_queue" breaks ordering when using maxConcurrentOperationCount=1 for operation queue


manuyavuz:
Hi, 

When an AFHTTPRequestOperation completes, it dispatches the task of calling success/failure blocks into `http_request_operation_processing_queue()` queue which is a concurrent queue by default. 

I need to assure that success/failure blocks of the requests will be called exactly the same order with operations' start order. I can guarantee the process order of requests via setting `requestManager.operationQueue.maxConcurrentOperationCount=1`. However, call order of success/failure blocks happens to be broken in edge cases because `http_request_operation_processing_queue()` works concurrently.

I can set `completionGroup` and `completionQueue` myself, but in this case in need to set `http_request_operation_processing_queue` which is currently not possible.

Is there any workaround for this?

Thanks

kcharwood:
Hi @manuyavuz,

Would you be able to create a test demonstrating the issue so I can look into it more? Thanks!

manuyavuz:
Hi @kcharwood,

I'll add two test cases one demonstrating the fail scenario, other fixes it via using a serial queue, and let you know.

Thanks

manuyavuz:
Hey @kcharwood,

I've added two test cases to #2935 demonstrating the issue. They pass on my machine (I'm using Xcode 7 GM and tried at least 40 times), however, one of them fails on travis CI. I can't catch the reason of failure, I guess it might result from travis using Xcode 6.4 (my guess).

Could you look for the tests and try to run on your machine? If it also fails on yours, I'll try to find the reason.

Thanks!

kcharwood:
closed via conversation in #2935

PR #2935, opened 2015-09-01
Ability to set a custom processing queue for AFHTTPRequestOperation

manuyavuz:
As stated in #2929, when an AFHTTPRequestOperation completes, it dispatches the task of calling success/failure blocks into `http_request_operation_processing_queue()` queue, which is a concurrent queue by default.

If somebody needs to assure that success/failure blocks of requests will be called exactly the same order with calling order of the requests, she can set `requestManager.operationQueue.maxConcurrentOperationCount = 1`. This will guarantee that requests will be sent to network layer in calling order, and their `completionBlock`s will be called in order also. However, because http_request_operation_processing_queue() works concurrently, the calling order of success/failure blocks happens to be broken in edge cases.

Setting a serial queue instance to this property will also gurantee ordering for success/failure callback blocks.

By the way, I'm not really sure about the name of the property, is there anyone having a better name proposal?

### DIFF
--- a/AFNetworking/AFHTTPRequestOperation.h
+++ b/AFNetworking/AFHTTPRequestOperation.h
@@ -65,6 +65,19 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setCompletionBlockWithSuccess:(nullable void (^)(AFHTTPRequestOperation *operation, id responseObject))success
                               failure:(nullable void (^)(AFHTTPRequestOperation *operation, NSError *error))failure;
 
+///---------------------------------
+/// @name Managing Callback Queues
+///---------------------------------
+
+/**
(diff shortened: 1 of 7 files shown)


kcharwood:
At the moment, I am going to leave this out. This seems like you are attempting to force some sort of synchronous networking, which we aren't specifically looking to support in AFNetworking.

We're also dropping support for `NSURLConnection` based API's, so I'm hesitant to add any new features here that will need to be maintained going forward.

If you haven't already, definitely look at moving over to the `NSURLSession` based API's!

jpmhouston:
Reason to do this: a response serializer that isn't re-entrant, eg. one using `XMLDictionaryParser`. Currently I'm having to use a do-nothing serializer (`return data;`) and do the actual decoding in the completion blocks. Note that I'm not a big fan of this workaround.

PR #2949, opened 2015-09-05
Remove single request

wscheper:
It would be nice if you had the ability to remove just one single cached request. Instead of needing to clear the whole cache.

Something like this.

### DIFF
--- a/UIKit+AFNetworking/UIImageView+AFNetworking.h
+++ b/UIKit+AFNetworking/UIImageView+AFNetworking.h
@@ -92,6 +92,13 @@ NS_ASSUME_NONNULL_BEGIN
        placeholderImage:(nullable UIImage *)placeholderImage;
 
 /**
+ Removes the cached request for the url from cache
+ 
+ @param url The url under with the request was initially created
+ */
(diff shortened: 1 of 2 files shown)


kcharwood:
Thanks for the PR!

There is actually a [little known feature](https://github.com/AFNetworking/AFNetworking/blob/master/UIKit%2BAFNetworking/UIImageView%2BAFNetworking.m#L194) in the image view category that already supports this!

If you create an `NSURLRequest` with a cache policy of `NSURLRequestReloadIgnoringCacheData` or `NSURLRequestReloadIgnoringLocalAndRemoteCacheData`, you can bypass the cache and force the request to go to the network.

wscheper:
That's not the same. Take an avatar image for instance. You want this to get persisted, but when the avatar gets updated, you may want to remove the older persisted image so you can retrieve the updated one. Right...?

kcharwood:
Depends on what you are trying to do I guess. Assuming your avatar is at the same URL, the method I listed above would bypass the cached the version and go to the network for it.

If the avatar URL is a different URL, then you wouldnt have the problem to being with. Am I misunderstanding your use case?

PR #2959, opened 2015-09-09
Capturing responseSerializer in AFURLSessionManagerTaskDelegate

OliverLetterer:
The NSURLConnection based AFHTTPRequestOperation had a way to override the AFHTTPRequestOperationManager's responseSerializer. Now that NSURLConnection is going away with iOS 9, I started migrating to AFHTTPSessionManager and find it extremely difficult to override the responseSerializer on a single, per request basis because AFURLSessionManagerTaskDelegate is capturing the sessionManager and grabbing it's responseSerializer to some unknown time on a private background queue.

I therefor made AFURLSessionManagerTaskDelegate capture the current responseSerializer which is backward compatible when the sessionManager's responseSerializer doesn't change. It's not backward compatible to the previous implementation when the sessionManager's responseSerializer does change, but the previous behavior was not deterministic.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -107,6 +107,7 @@ typedef void (^AFURLSessionTaskCompletionHandler)(NSURLResponse *response, id re
 
 @interface AFURLSessionManagerTaskDelegate : NSObject <NSURLSessionTaskDelegate, NSURLSessionDataDelegate, NSURLSessionDownloadDelegate>
 @property (nonatomic, weak) AFURLSessionManager *manager;
+@property (nonatomic, strong) id<AFURLResponseSerialization> responseSerializer;
 @property (nonatomic, strong) NSMutableData *mutableData;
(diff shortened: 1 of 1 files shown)


OliverLetterer:
It seems #1723 already implemented this once Oo I have no idea if this behavior has been removed on purpose or not.

kcharwood:
Long time no see @OliverLetterer! :beers:

Can you let me know how `AFCompoundResponseSerializer` misses on your specific problem here?

kcharwood:
ping @OliverLetterer

OliverLetterer:
@kcharwood Somehow, your response got lost in my inbox :S Sorry I have kept you waiting.

I have been able to integrate AFCompoundResponseSerializer in my use case :beers: but I could imagine inconsistent web services where this approach might not work as flawless. In general, this might boil down to a more philosophical discussion about whether the dependency on shared mutable state is good or bad. Since it doesn't look like anyone has hit this issue before and I'm not that versed in the AFNetworking source anymore, I'm going to close this issue 😊

PR #2967, opened 2015-09-12
Add TARGET_OS_TV conditionals

mhpavl:
go get the project to build

### DIFF
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -23,7 +23,7 @@
 
 #import <AssertMacros.h>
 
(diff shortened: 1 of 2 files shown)


drekka:
I've been trying to do this but it won't compile. I've setup a tvOS target and tried to compile it. But I get a whole set of errors around UI components that are not available in tvOS. Specifically UIAlertView.

kcharwood:
We also need to update the podspec to define `TARGET_OS_TV` if it isnt defined

kcharwood:
We'll keep [this branch](https://github.com/AFNetworking/AFNetworking/tree/tvOS) up to date until Xcode 7.1 hits GM.

Thanks!

Issue #2805, opened 2015-06-23
task.taskDescription 


marinotws:
Hi, following several upload in backgrond, when my app is closed and re-opened by the system to new data going to read the task.taskDescription but the content stored disappeared ... where am I wrong?

    setTaskDidCompleteBlock:^(NSURLSession *session, NSURLSessionTask *task, NSError *error) {
    .....
    NSString *param = task.taskDescription;
    
thanks for the help !

marinotws:
![schermata 2015-06-23 alle 11 15 54](https://cloud.githubusercontent.com/assets/8616947/8304030/c96b28be-19a2-11e5-9a4b-b84ac6b2ef01.png)

this is the result ... I'm pretty sure that the problem is after [task suspend] and then [task resume] ....

marinotws:
please @mattt help ! :-)

kcharwood:
I'm not exactly sure what you are asking here. Can you be more specific, or provide a clearer example on what you expect the output to be?

joshsloat:
I think I have some additional insights that could help to shine a light on the problem.

There was a fix introduced by @tangphillip around tasks and network activity in the following PR: #2477. Of particular note:

```uploadTask.taskDescription = [self descriptionForAllTasks];```

This fix is currently causing some side effects. By setting taskDescription in this manner, when being resurrected from the background, the change is over-writing what the caller may have initially specified for the taskDescription. For example, let’s say I:
 
1. Create an upload task with a taskDescription of “TaskA” (so that I can make conditional “which task was it” decisions in callbacks since I am chaining together several dependent tasks).
1. Kick off my task, crash my app and then relaunch the app. 
1. At that point, I should be able to re-connect to my session and use the taskDescriptions I specified pre-crash (which are persisted properly with the task) to take some specific action if ```[task.taskDescription isEqualToString:@“TaskA”]```.

With the approach implemented in the referenced PR, upon resurrection, “TaskA” is being overwritten with a memory address that is of no use to me and no reliable way of knowing which task finished, reported progress, etc.

marinotws:
sorry for my absence ... @joshsloat He has fully exposed the problem !

sust86:
I'm currently running into the same issue. I need the taskDescription property to map downloadTasks when the application returns.
As far as I see AFNetworking is using taskDescription somehow for the loading indicator observer.
But there is no other way to give background download tasks some payload to identify them later again. 

Is there some alternative option we have?

Thanks

[...]

PR #2477, opened 2014-12-17
Network indicator no longer spins forever when using >1 session manager

tangphillip:
Issue details at #2476. Now matches tasks with `AFURLSessionManager`s by setting all tasks' `taskIdentifier`s to the stringified address of the manager.

Issue #2476, opened 2014-12-17
Multiple AFURLSessionManagers + AFNetworkActivityIndicatorManager = never-ending activity indicator


tangphillip:
When using multiple `AFURLSessionManager`s, AFNetworking generates many extraneous `AFNetworkingTaskDidResumeNotification` notifications, causing the `AFNetworkActivityIndicatorManager` to show the activity indicator permanently.

Sample project demonstrating the issue, courtesy of @eytanbiala: https://dl.dropboxusercontent.com/u/3067376/AFNetworkingTest-2458.zip

**Root cause:** `AFURLSessionManager`s were determining that tasks belong to them...when they really don't. Managers were using, "Do I have a task with this `taskIdentifier`?" as shorthand for, "Was this task created by me?" Unfortunately, `taskIdentifier`s are unique only within their `NSURLSession`, and therefore any given identifier can be (and frequently is) used in multiple sessions at the same time.

phoney:
I have also run into this bug (and spent an inordinate amount of time tracking it down). 

Your fix gets to the heart of the matter but I don't think it's a good idea to commandeer the taskDescription for this purpose. This will interfere with developers using the taskDescription for their own purposes. I use the taskDescription in my own code for background download tasks. I save the file path where the file is to be saved in the taskDescription.

The underlying bug is due to the taskIdentifier not being unique across the entire app but only across each session manager, as you say. Therefore it was a mistake to use the taskIdentifier as the key for the mutableTaskDelegatesKeyedByTaskIdentifier dictionary. So another simple fix is to change the key from the taskIdentifier to the task pointer. Something like this

    delegate = self.mutableTaskDelegatesKeyedByTaskIdentifier[@((NSUInteger) task)];

in three places fixes the bug as well. I guess you'll want to change the name of the dictionary too. This would be almost the same as your fix but would avoid using the taskDescription.

tangphillip:
If we can rely on `NSURLSessionTask` having a consistent pointer throughout its life, that's a good solution.

`NSURLSessionTask` conforms to `NSCopying`. What would happen if I made a `copy` of a task? I don't know enough about `NSURLSession` to decide if that's a thing worth supporting.

phoney:
I don't think copying is an issue. Yes, a copy would get a new address. But a copy wouldn't belong to our AFURLSessionManager.

Other solutions that come to mind are 
to simply keep a list of active session tasks inside each AFURLSessionManger. 
[...]

PR #2973, opened 2015-09-13
Let app devs use taskDescription by using associated objects internally

tangphillip:
A _while_ ago, we started using a swizzle-observation of `-[NSURLSessionTask resume]` and `-[NSURLSessionTask suspend]` instead of KVO-observing the `state`, due to a crashing bug inside KVO. For background, see #1477.

Because app developers can use multiple instances of `AFURLSessionManager`, and because swizzle-observation will make every session manager see every task's state changes, regardless of which manager actually owns the task, we needed a way to associate the task with the session manager (#2476). We ended up using the `NSURLSessionTask`'s `taskDescription` for that purpose (#2477).

But some app developers have expressed interest in using the `taskDescription` for their own purposes (#2805), so this pull request replaces that implementation with one using `objc_(get|set)AssociatedObject`. We're using the runtime already...why not? (Sigh.)

Addresses #2805.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -282,6 +282,7 @@ static inline BOOL af_addMethod(Class class, SEL selector, Method method) {
     return class_addMethod(class, selector,  method_getImplementation(method),  method_getTypeEncoding(method));
 }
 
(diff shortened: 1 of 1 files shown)


tangphillip:
(This doesn't actually work. I'll reopen it when it does!)

PR #2974, opened 2015-09-13
Let app devs use taskDescription by using associated objects internally

tangphillip:
A _while_ ago, we started using a swizzle-observation of `-[NSURLSessionTask resume]` and `-[NSURLSessionTask suspend]` instead of KVO-observing the `state`, due to a crashing bug inside KVO. For background, see #1477.

Because app developers can use multiple instances of `AFURLSessionManager`, and because swizzle-observation will make every session manager see every task's state changes, regardless of which manager actually owns the task, we needed a way to associate the task with the session manager (#2476). We ended up using the `NSURLSessionTask`'s `taskDescription` for that purpose (#2477).

But some app developers have expressed interest in using the `taskDescription` for their own purposes (#2805), so this pull request replaces that implementation with one using `objc_(get|set)AssociatedObject`. We're using the runtime already...why not? (Sigh.)

Addresses #2805.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -282,6 +282,7 @@ static inline BOOL af_addMethod(Class class, SEL selector, Method method) {
     return class_addMethod(class, selector,  method_getImplementation(method),  method_getTypeEncoding(method));
 }
 
+static void *AFNSURLSessionTaskSessionManagerKey = &AFNSURLSessionTaskSessionManagerKey;
 static NSString * const AFNSURLSessionTaskDidResumeNotification  = @"com.alamofire.networking.nsurlsessiontask.resume";
 static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofire.networking.nsurlsessiontask.suspend";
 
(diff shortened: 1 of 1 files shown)


kcharwood:
Would this work if a task was restored from the background?

kcharwood:
Could you put together some tests for this issue? Would be a good candidate to move into the 3.0.0 branch.

kcharwood:
I refactored just a bit and added some tests in d284c551. Let me know what you think!

tangphillip:
Hey! I think your refactor looks great! Sorry I haven't responded; I haven't had time to put anything together. Regarding restore from background: I'm not sure. I didn't test that case.

kcharwood:
I'm actually worried this will break background downloads... Will need to try to put together a test app and try it out.

hhanesand:
This is something that I'm interested in if it makes any difference, I would like to use the `taskDescription` for debugging purposes.

joshsloat:
FWIW, I implemented this fix almost exactly in a test brach of mine to address what I mention in #2805. I can't speak to the other concerns mentioned here, but this solution certainly does resolve the issue of  not being able to reliably use task descriptions upon app resurrection.

PR #2981, opened 2015-09-15
Fixed overwrite a file

iKorich:
Hi,

You have problems in AFURLSessionManager
You can't overwrite a file if it exist

This can move file if it doesn't exist:
```
[[NSFileManager defaultManager] moveItemAtURL: toURL: error:];
```

Test:
1. Start download a file
2. Update file on server
3. Start download a file
4. Check file


kcharwood:
Thanks for the PR. Would you be able to add a test?

iKorich:
Sorry, I haven't time to do a test

iKorich:
Moved "if" to new line

kcharwood:
Is there any backwards compatibility issues to think about here? Is overwriting the correct, default choice?

AFNetworking/AFURLSessionManager.h
```diff
@@ -444,6 +444,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ Rewriting `-[NSFileManager moveItemAtURL:toURL:error:] to allow replace old data.
+ 
+ @param srcURL A path of current file url. Must not be `nil`.
+ @param dstURL A file destination. Must not be `nil`.
+ @param downloadTask The session download task..
+ */
++ (void)moveItemAtURL:(NSURL *)srcURL toURL:(NSURL *)dstURL downloadTask:(NSURLSessionDownloadTask *)downloadTask;
+
 @end
```
kcharwood: What is the need to have this be public?

iKorich:
> Is there any backwards compatibility issues to think about here?
Yes, because it's a bug

> Is overwriting the correct, default choice?
Yes, because user can't download a updated file

> What is the need to have this be public?
I can make it static, because it using AFURLSessionManagerTaskDelegate and AFURLSessionManager

kcharwood:
@iKorich would you be able to target this at the 3_0_0_branch ? Would like to pull it in there.

kcharwood:
If this condition is met, `AFURLSessionDownloadTaskDidFailToMoveFileNotification` is posted with the download task, and the user info of the error, which contains this information:

```
{
    NSDestinationFilePath = "/Users/kevin.harwood/Library/Developer/CoreSimulator/Devices/37204238-3A01-41B4-BAFA-02C3E276B372/data/Documents/image.png";
    NSFilePath = "/Users/kevin.harwood/Library/Developer/CoreSimulator/Devices/37204238-3A01-41B4-BAFA-02C3E276B372/data/tmp/CFNetworkDownload_eWgL8c.tmp";
    NSSourceFilePathErrorKey = "/Users/kevin.harwood/Library/Developer/CoreSimulator/Devices/37204238-3A01-41B4-BAFA-02C3E276B372/data/tmp/CFNetworkDownload_eWgL8c.tmp";
    NSUnderlyingError = "Error Domain=NSPOSIXErrorDomain Code=17 \"File exists\"";
    NSUserStringVariant =     (
        Move
    );
}
```

In addition, in the destination block you would have the ability to check to see if an item is at the path you are going to return, giving you a chance to remove it if necessary.

Given the hooks available, I would err on the side of leaving the existing file in place. I would hate to automatically delete a file if the user provided the wrong path.

Thoughts?

kcharwood:
@iKorich any thoughts on my above comment?

iKorich:
Hi @kcharwood

I check it, and AFURLSessionDownloadTaskDidFailToMoveFileNotification is too late to fix a problem

also user can remove file in downloadTaskDidFinishDownloading block
but he should add a:
```
NSFileManager *fm = [NSFileManager defaultManager];
            if([fm fileExistsAtPath:[self.downloadFileURL path]]) {
                [fm removeItemAtURL:self.downloadFileURL error:&error];
            }
```
to many places in code

may be add some another block?

```
[self.manager setRewriteDownloadedFileBlock:^NSURL *(NSURLSession *session, NSURLSessionDownloadTask *downloadTask, NSURL *location) {
        return YES;
    }];
```

and user can choice file types to overwrite it or no

kcharwood:
@iKorich I believe the `destination` block would give you a chance to check that path, and delete a file if its there, before returning the path?

iKorich:
@kcharwood Yes

PS. In our project we have few download managers(for different servers) and that is not comfortable to put one code (remove file) to all places
And we always need get latest configs and resources

kcharwood:
@iKorich At the moment, I would err on _not_ automatically deleting the file, simply because I'm not sure what other use cases out there that might effect. The destination block gives you a hook to check the file path you are going to return, and do any action you need to do to make sure you are returning a file URL that does not have an item at that destination. Given that, I would say stick with that approach for now.

@cnoon @0xced @jshier Any thoughts on the sensible default here? Am I being to conservative?

cnoon:
I'm with you on this one @kcharwood...stay conservative. The goal here of AFN is to download the file to a specific location, not add a bunch of other hooks to control the file system manipulation and/or possibly error handling. That is all client code that should be handled by the client IMO.

As you have mentioned, in the event that the file already exists, the client needs to make a conscious decision as to when to purge the previous file. When that should happen is completely up to the client. It may be able to happen before the request starts, right before a completed request copies the file or even needs to be put into a staging directory before swapping the files. If anything goes wrong in any of these scenarios, the client needs to have full control to handle the error conditions. If we bury this automatic logic inside AFN, client's won't be able to handle the error logic properly because it will most likely get buried.

As I said, stay conservative. 👍🏼

kcharwood:
Thanks @cnoon 

I'm going to go ahead and close this one out then. Would be happy to hear anyone else with strong opinions and use cases on a better solution for this.

PR #2984, opened 2015-09-16
String representation for boolean values in query string

matteonovelli:
I have the same issue #931 and I think it's convenient to have the option to parametrize the desired boolean representation of boolean values.


kcharwood:
As stated in #931, I'd like to leave this out of the core. I'd recommend subclassing a serializer to get this specific behavior for your application.

:beers:

matteonovelli:
Thanks for the response. I think is not possible to get this behavior by subclassing the serializer because AFQueryStringPair is a private class. Also I added a property on the serializer to parametrize the behavior.

PR #2996, opened 2015-09-22
[*] responseSerializer using fix

nahashsvb:
Better to use AFCompoundResponseSerializer  instead AFJSONResponsSerializer, because we do not know what type of response the server sends.

### DIFF
--- a/AFNetworking/AFHTTPRequestOperationManager.m
+++ b/AFNetworking/AFHTTPRequestOperationManager.m
@@ -50,25 +50,31 @@
     if (!self) {
         return nil;
     }
-
+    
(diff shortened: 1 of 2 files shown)


kcharwood:
Interesting change. Hadn't thought about this before. Since JSON is by the far the most used type of serializer, I'm hesitant to add this in simply because so many projects don't need it, and I don't think its particularly challenging to change the response serializer if you have a different format. In fact, I like developers needing to do that step because I think it opens up possibilities for them to recognize they can create custom serializers to do their own parsing.

For now, I'm going to leave this out, unless we see this as a big problem for the user base.

Thanks for the taking the time to put to this together!

:beers:

nahashsvb:
Thank you considered my offer.
Best regards!

PR #3001, opened 2015-09-23
add allowExpiredAuthorizedCertificates support

dake:


kcharwood:
Thanks for the PR.

Since this code touches the security policies, can you please provide tests demonstrating the failing scenario, and how this patch causes it to pass, as well as tests demonstrating the negative cases?

We want to get as much coverage as possible with any code touching the security policies.

kcharwood:
I'm hesitant on this one as well. I would consider it best practice to not have this enabled, which I feel would encourage people to work around a bad practice.

I've leave this open for community :+1: interest

Thanks for taking the time to put together a PR!

kcharwood:
Would also like @cnoon @jshier @kylef opinion here.

cnoon:
This concerns me. Why would we need an `allowExpiredAuthorizedCertificates` in addition to the already existing `allowInvalidCertificates` flag? Seems like this is a PR to account for something that already existed, just named something else.

dake:
@cnoon `allowExpiredAuthorizedCertificates` means some recoverable certificate errors, like expired certificate, but `allowInvalidCertificates` contains invalid certificates, which may be unrecoverable.

cnoon:
So after digging through the docs @dake, I don't think it's safe to make that assumption that `allowExpiredAuthorizedCertificates` is ONLY going to operate on expired certificates. The docs simply use expired certificates as an example. 

> *kSecTrustResultRecoverableTrustFailure*: Trust denied; retry after changing settings. For example, if trust is denied because the certificate has expired, you can ask the user whether to trust the certificate anyway. If the user answers yes, then use the SecTrustSettingsSetTrustSettings function to set the user trust setting to kSecTrustResultProceed and call SecTrustEvaluate again. This value may be returned by the SecTrustEvaluate function but not stored as part of the user trust settings.

They don't explicitly list the conditions in which the `kSecTrustResultRecoverableTrustFailure` type will actually be returned. Therefore, IMO, I don't think it's a very wise decision to set this new `allowExpiredAuthorizedCertificates` to `true` without knowing EXACTLY what it is you are enabling.

I also find it extremely confusing as a user to distinguish between these two property names. It is very important to have the properties controlling the certificates evaluation be extremely clear.

dake:
@cnoon I really agree with you, `allowExpiredAuthorizedCertificates` is ambiguious. I was hesitated before, but I would remove it from code. thanks so much.

kcharwood:
I agree with everything you said there @cnoon. I'm going to close this one out.

Thanks again everyone! 

:beers:

Issue #3014, opened 2015-09-29
Creating multiple instances of AFHTTPRequestOperationManager causes reachability block to be overriden


mobitar:
When you create an instance of `AFHTTPRequestOperationManager`, in the `initWithBaseURL:` initializer, the operation manager is assigned the shared `AFNetworkReachabilityManager`:

```objective-c
self.reachabilityManager = [AFNetworkReachabilityManager sharedManager];
```

This means if your app has multiple `AFHTTPRequestOperationManager`'s, setting the `networkReachabilityStatusBlock` on the `reachabilityManager` property of the operation manager instance overrides the block you may have set on other instances of an operation manager's `reachabilityManager`.

This is undesired behavior, as ultimately, only the last block set will be executed in the event of a reachability change.

The solution would be to assign each `AFHTTPRequestOperationManager` its own unique instance of `AFNetworkReachabilityManager`, rather than using the `sharedManager`.

Possible solution: https://github.com/mobitar/AFNetworking/commit/4829c0b012f36c280efca135eaf4d4e782426be0

tommypeps:
I think this doesn't a bug, really Singleton implementation define a unique instance for all the execution.
If you call the singleton on 2 instance different normally one must disappear.

If you want, instances 

```Objective-C
self.reachabilityManager = [AFNetworkReachabilityManager sharedManager];
```
in appDelegate:
```Objective-C
self.reachabilityManager = [AFNetworkReachabilityManager sharedManager];
[self.reachabulityManager startMonitoring];
.
.
.
[self.reachabilityManagersetReachabilityStatusChangeBlock:^(AFNetworkReachabilityStatus status) {
   [[NSNotificationCenter defaultCenter] postNotificationName:@"newConection"
                                                                             object: @{@"status":AFNetworkReachabilityStatus}];
}]
```
:arrow_up:  Send notification at your other class for know when arrive conection




other Option, implemented only a instance but not Singleton
```Objective-C
self.reachabilityManager =  [AFNetworkReachabilityManager managerForDomain:@"my domain"];
    [self.sharedManagerAF startMonitoring];
.
.
.
[self.reachabilityManagersetReachabilityStatusChangeBlock:^(AFNetworkReachabilityStatus status) {
//TODO:
}]
```

kcharwood:
The AF Documentation does state:

>The network reachability manager. `AFURLSessionManager` uses the `sharedManager` by default.

[...]

PR #3015, opened 2015-09-29
Fix for issue #3014: unique instances of reachability manager for AFHTTPRequestOperationManager

mobitar:


kcharwood:
From @mobitar in #3014:

When you create an instance of `AFHTTPRequestOperationManager`, in the `initWithBaseURL:` initializer, the operation manager is assigned the shared `AFNetworkReachabilityManager`:

```objective-c
self.reachabilityManager = [AFNetworkReachabilityManager sharedManager];
```

This means if your app has multiple `AFHTTPRequestOperationManager`'s, setting the `networkReachabilityStatusBlock` on the `reachabilityManager` property of the operation manager instance overrides the block you may have set on other instances of an operation manager's `reachabilityManager`.

This is undesired behavior, as ultimately, only the last block set will be executed in the event of a reachability change.

The solution would be to assign each `AFHTTPRequestOperationManager` its own unique instance of `AFNetworkReachabilityManager`, rather than using the `sharedManager`.

Possible solution: https://github.com/mobitar/AFNetworking/commit/4829c0b012f36c280efca135eaf4d4e782426be0

pronebird:
@kcharwood any follow up on this. @mobitar's patch seems nice.

kcharwood:
At the risk of any potential backwards compatibility issues, I'm going to push this to 3.0.0. #3111 

Thanks! I'll get it merged in soon.

PR #3023, opened 2015-10-05
Make sure NSNotificationCenter interactions take place in their own queue

### LINKED ISSUE #2439: AFNetworkActivityIndicatorManager taskDidResume erratic crash
Hi Matt!

Firstly, many thanks for your hard work in AFN and I've done my best to make sure the issue isn't on my end!

I'm using a UIRefreshControl in a UITableViewController similar to how GlobalTimelineViewController uses one in the AFNetworking example.  However, after 3-5 table refreshes, I'm getting a unrecognized selector crash when AFURLSessionManager's  `taskDidResume:` (AFURLSessionManager.m:402) is trying to send a `AFNetworkingTaskDidResumeNotification` to `networkRequestDidStart:` of an instance of AFNetworkActivityIndicatorManager.  

This issue occurred in rev 29511 and still occurs in the latest  29542, so this is not newly introduced.

Separately, I wanted to point out that the property `refreshControl` is redundant in GlobalTimelineViewController.m:34 since it's also available in UITableViewController.

Thanks,
Drew

sergiou87:
This should fix #2439 

If I'm right, the problem was this:
 1. Thread A: `- [NSURLSessionDataTask resume]` is called. The swizzled method `af_resume` posts a `AFNSURLSessionTaskDidResumeNotification` notification. The callback for that notification is not invoked (yet).
 2. Thread B: an instance of `AFHTTPSessionManager` is deallocated. That means, it removes itself as observer of `AFNSURLSessionTaskDidResumeNotification` (but now it's too late!).
 3. Thread A: the callback of `AFNSURLSessionTaskDidResumeNotification` is invoked (`-[AFURLSessionManager taskDidResume:]`) in the instance deallocated in thread B (step 2) and tries to access a property of it.
 4. CRASH!!!

You can reproduce it with the following code in your AppDelegate (you may need to run it several times, and each time you will get different crashes depending on when it switches between threads). This doesn't happen if you built the app with iOS 9 SDK and running on iOS 9, because of the changes Apple introduced on `NSNotificationCenter`, but it definitely happens on iOS 8 (no matter what SDK you used) and iOS 9 if the app was built with iOS 8 SDK.

The numbers for the iterations (and even part of the code) is kind of random, I just wanted to copy the exact code that reproduced the problem. It just creates a bunch of `AFHTTPSessionManager` with short life-span, and concurrently creates tasks and resumes them.

```objective-c
#import "AFNetworking.h"

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    AFHTTPSessionManager *sm = [[AFHTTPSessionManager alloc] initWithBaseURL:[NSURL URLWithString:@"http://127.0.0.1:8080"]];

    for (NSUInteger k=0; k < 20; k++)
    {
        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
            for (NSUInteger i=0; i < 100; i++)
            {
                NSURLSessionDataTask *task = [sm dataTaskWithRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:@"http://127.0.0.1:8080"]] completionHandler:nil];
                [task resume];

            }
        });

        for (NSUInteger i=0; i < 100; i++)
        {
            @autoreleasepool {
                AFHTTPSessionManager *j = [[AFHTTPSessionManager alloc] initWithBaseURL:[NSURL URLWithString:@"http://127.0.0.1:8080"]];

                [j invalidateSessionCancelingTasks:YES];
            }
        }
    }

    return YES;
}
```

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -24,6 +24,16 @@
 
 #if (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000) || (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090)
 
+static dispatch_queue_t url_session_manager_notification_queue() {
+    static dispatch_queue_t af_url_session_manager_notification_queue;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        af_url_session_manager_notification_queue = dispatch_queue_create("com.alamofire.networking.session.manager.notification", DISPATCH_QUEUE_SERIAL);
+    });
+
+    return af_url_session_manager_notification_queue;
+}
+
 static dispatch_queue_t url_session_manager_creation_queue() {
     static dispatch_queue_t af_url_session_manager_creation_queue;
     static dispatch_once_t onceToken;
@@ -368,7 +378,9 @@ static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofi
     [self af_resume];
     
     if (state != NSURLSessionTaskStateRunning) {
-        [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidResumeNotification object:self];
+        dispatch_async(url_session_manager_notification_queue(), ^{
+            [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidResumeNotification object:self];
+        });
     }
 }
 
@@ -378,7 +390,9 @@ static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofi
     [self af_suspend];
     
     if (state != NSURLSessionTaskStateSuspended) {
-        [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidSuspendNotification object:self];
+        dispatch_async(url_session_manager_notification_queue(), ^{
+            [[NSNotificationCenter defaultCenter] postNotificationName:AFNSURLSessionTaskDidSuspendNotification object:self];
+        });
     }
 }
 @end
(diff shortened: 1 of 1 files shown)


sergiou87:
I made some changes and broke the tests, I'll fixe them ASAP.

kcharwood:
It is possible to write a unit test displaying this behavior?

sergiou87:
I can't think of a deterministic way to reproduce this: you need to post the notification in one thread, and right after that and before it finishes executing the callback, the `AFURLSessionManager` instance must be deallocated. That's why my sample code creates lots of instances (that are almost immediately deallocated) and posts lots of notifications concurrently, and even that is not 100% reliable (although for me it crashed most of the times I ran it).

Anyway, I'm open to suggestions :sweat_smile:

kcharwood:
@sergiou87 What version of Xcode and iOS were you reproducing this on? Trying now.

kcharwood:
This is definitely reproducible in the app delegate. I can get a test target to crash though.

kcharwood:
I'm also having trouble getting the AFURLSessionManagerTests to pass locally. Are you using any issues there?

sergiou87:
Hey! @kcharwood : we get thousands of those crashes in a released app :disappointed: 

Regarding the tests, all the asynchronous tests weren't needed because the suspend/resume notifications were posted synchronously. Now all tests run in parallel sharing the same sync queue for those notifications (`url_session_manager_notification_queue`), so they block each other and that makes the tests time out. If you increase the test timeouts to a high value (10 was enough on my laptop) it works, but of course that's not the best way to fix it…

We should be somehow able to inject different queues for each test so they continue being independent while on release they all keep the same queue, but I hadn't time to work on that, sorry :confused: 

And again, I'm open to suggestions :stuck_out_tongue_winking_eye:

kcharwood:
I wonder if this could be filtered down a different way? Maybe only adding the notification observer for the specific tasks a session creates? Then it wouldn't get notified from tasks on other sessions?

kcharwood:
@sergiou87 what do you think about 8370ccb? All tests pass with that patch. Can you see if that makes a difference in your code base?

Thoughts from @cnoon @kylef @jshier ?

I'm assuming `object` on `NSNotificationCenter addObserver:selector:name:object` is a weakly held reference and I don't need to explicitly unregister it, but I can't find documentation backing that up.

jshier:
Unless you're running on iOS 9 only, `NSNotificationCenter` observers need to be explicitly removed using `removeObserver:`.

kcharwood:
Hey @jshier. Not the observer, but the object.

jshier:
Sorry, missed that part.

For the purposes of memory management that's correct, AFAICT. However, the `NSNotificationCenter` docs do state that `removeObserver:` should be called before `object:` is deallocated. I'm not sure that possible in the alternate solution.

kcharwood:
If that's the case, we would need to remove the observer in the completion blocks for all the tasks.

sergiou87:
@kcharwood Sorry I've been too busy these days… the problem with that alternate solution is that it doesn't solve the problem, it just makes it less probable (because the session manager doesn't listen to EVERY task in the app anymore), but it still can happen that the session manager gets deallocated right after a task posted a notification and before it's handled.

So you're patch doesn't fix the problem, but I'd keep it because IMHO it's a great addition.

sergiou87:
@kcharwood  Actually, maybe with that change you made we can switch to one notification queue per session manager, instead of having one shared which would make the tests independent again.

kcharwood:
@sergiou87 It may actually fix it though. Is it possible for a session to be dealloc'ed if existing tasks associated with it are still in flight? If thats not possible, then this just might solve it. I need to look into some the cancel state today and verify those flows.

kcharwood:
I made some additional updates in 69f7ca0 to manage adding/removing observers.

sergiou87:
> It may actually fix it though. Is it possible for a session to be dealloc'ed if existing tasks associated with it are still in flight? If thats not possible, then this just might solve it. I need to look into some the cancel state today and verify those flows.

That's definitely a good point :smile: My knowledge about it is very scarce, but I don't see anything obvious that points in that direction, it seems like tasks can outlive sessions… but maybe I'm missing something :confused:

kcharwood:
I think I'm going to move forward with #3152 for now. Thanks for digging in on this!

:beers:

sergiou87:
Awesome! Nice work! And sorry I couldn't be of more help :sweat_smile:

PR #3028, opened 2015-10-07
crash fix with iOS 7 stringByAddingPercentEncodingWithAllowedCharacters method

softenhard:


kcharwood:
Thanks @softenhard. This has been on my list to fix, but I hadn't yet thought about how to handle some of the double byte characters like emoji. I know relying on NSString length and enumeration can result in a split character. Any thoughts on that?

softenhard:
@kcharwood Oh yes, forgot about that while fast-fixing my project :)
So I've added a couple of lines to support multibyte characters.
Check it out:
```
static NSUInteger const batchSize = 50;

	NSInteger index = 0;
	NSMutableString *escaped = @"".mutableCopy;

	while (index < string.length) {
		NSUInteger length = MIN(string.length - index, batchSize);
		NSRange range = NSMakeRange(index, length);

		// To avoid breaking up character sequences such as 👴🏿
		range = [string rangeOfComposedCharacterSequencesForRange:range];

		NSString *substring = [string substringWithRange:range];
		NSString *encoded = [substring stringByAddingPercentEncodingWithAllowedCharacters:allowedCharacterSet];
		[escaped appendString:encoded];

		index += range.length;
	}
```

kcharwood:
What happens if the split is on the mutlibyte character? Is the character included in both results?

Could you add some unit tests demonstrating the fix? Thanks!

kcharwood:
Updated in 6b6d41d

kcharwood:
Merged with 6b6d41d

kcharwood:
Merged in 3_0_0 with 039a317

Coeur:
As AFNetworking code is referencing this issue in https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFURLRequestSerialization.m#L54, I'd like to quote in this pull request the details of this issue.

The reason and the workaround was discovered by @PrideChung and is a memory crash issue in `stringByAddingPercentEncodingWithAllowedCharacters`, reported in https://github.com/Alamofire/Alamofire/issues/206. Scope of this issue was refined by @cnoon as only affecting iOS7 and iOS8, not affecting iOS9:
> Batching is required for escaping due to an internal bug in iOS 8.1 and 8.2. Encoding more than a few hundred Chinese characters causes various malloc error crashes. To avoid this issue until iOS 8 is no longer supported, batching MUST be used for encoding. This introduces roughly a 20% overhead.

CodeLife2012:
nice

fighterLS:
* | Application Specific Information: *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSCFString appendString:]: nil argument'
  -- | --
  0 | CoreFoundation | 0x0000000202a74ec4 ___exceptionPreprocess + 228
  1 | libobjc.A.dylib | 0x0000000201c45a40 _objc_exception_throw + 56
  2 | CoreFoundation | 0x000000020297b594 +[NSException raise:format:] + 116
  3 | CoreFoundation | 0x00000002029ea7d8 _mutateError + 204
  4 | homework | 0x0000000102ae95dc AFPercentEscapedStringFromString (AFURLRequestSerialization.m:73)
  5 | homework | 0x0000000102ae9880 -[AFQueryStringPair URLEncodedStringValue] (AFURLRequestSerialization.m:108)
  6 | homework | 0x0000000102ae9a8c AFQueryStringFromParameters (AFURLRequestSerialization.m:0)
  7 | homework | 0x0000000102aec560 -[AFHTTPRequestSerializer requestBySerializingRequest:withParameters:error:] (AFURLRequestSerialization.m:0)
  8 | homework | 0x0000000102aeb878 -[AFHTTPRequestSerializer requestWithMethod:URLString:parameters:error:] (AFURLRequestSerialization.m:376)

PR #3059, opened 2015-10-14
Fixed race condition in Reachability

ejud:
If a reachability update occurs at the right time after calling `-startMonitoring`, the `AFNetworkReachabilityCallback` function can be called before the first-time reachability update is posted. Therefore, a race condition existed that could cause the incorrect reachability status to be reported.

This change works around the issue by tracking whether the callback has yet been invoked, and if so, balking during the first-time reachability update. 

I tested this with an iPhone 6 running iOS 9.0.2, and toggling Airplane Mode during app launch (when we call `-startMonitoring`).

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -188,8 +188,12 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
         return;
     }
 
+    __block BOOL receivedReachabilityUpdate = NO;
+
     __weak __typeof(self)weakSelf = self;
     AFNetworkReachabilityStatusBlock callback = ^(AFNetworkReachabilityStatus status) {
+        receivedReachabilityUpdate = YES;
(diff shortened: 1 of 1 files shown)


kcharwood:
Can you think of anyway to test this?

ejud:
I reproduced this with the following steps:
- Create an empty iOS project
- in `application:didFinishLaunchingWithOptions:`, add `sleep(5)` before returning.
- in the main view controller, set up the reachability change callback during `init`, and write changes to a text view and/or the console.
- Launch the app. While it is launching (within the first 5 seconds), use the Control Center to enable Airplane Mode.
- You'll see that the most recent reported reachability status is "reachable", which is incorrect.

For a sample project, see https://github.com/ejud/AFNetworkingEdgeCaseTest

kcharwood:
@ejud  what do you think about #3117?

kcharwood:
I've merged in #3117.

Thanks!

PR #2930, opened 2015-08-31
use OS_OBJECT_USE_OBJC instead of OS_OBJECT_HAVE_OBJC_SUPPORT

canius:
When using cocoapods, we need fix this to avoid compile error.
See http://stackoverflow.com/questions/27267865/strong-dispatch-queue-t-in-cocoapods-library

PR #3065, opened 2015-10-16
Switch to OS_OBJECT_USE_OBJC

kylef:
As per #2930, this won't build for watchOS 2 otherwise.

I know this is fixed for AFNetworking 3 since we dropped older OS supports, I'd suggest we add this for the next patch/maintenance release.

/cc @kcharwood


kcharwood:
:+1:

PR #3075, opened 2015-10-19
about notification center 

coderyi:
about notification center because they use postNotification or addObserver ，so i think when their object is dealloc,they will remove


kcharwood:
`AFNetworkReachabilityManager` does not add it self to the Notification Center, so there is no need to remove it.

PR #3077, opened 2015-10-19
add support for the content-type of text/plain

RolandSung:


kcharwood:
`text/plain` is not always JSON, therefore we can't take this patch. If you have a server that is returning JSON with this content-type, you can add support for it by updating the `acceptableContentTypes` property on the response serializer.

:beers:

PR #3085, opened 2015-10-21
AFNetworkActivityIndicatorManager visibilityDelay property

lubiluk:
A property called visibilityDelay added to
AFNetworkActivityIndicatorManager. It controls the time to wait before
showing the network activity indicator.  This is added to be able to
prevent showing the activity indicator for short periods of time as
suggest iOS Human Interface Guidelines.
https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/MobileHIG/Controls.html#//apple_ref/doc/uid/TP40006556-CH15-SW44


kcharwood:
Thanks for the PR! We actually avoid flickering by having the indicator remain on the screen a little while longer. You can see the code [here](https://github.com/AFNetworking/AFNetworking/blob/master/UIKit%2BAFNetworking/AFNetworkActivityIndicatorManager.m#L100).

:beers:

PR #3086, opened 2015-10-21
Network activity indicator delay

lubiluk:
When `visibilityDelay` is used then the default delay is not necessary.
This allows greater control over when the indicator should show.

It relates to this pull request: https://github.com/AFNetworking/AFNetworking/pull/3085.
Only this time it turns off the default 0.17 delay when `visibilityDelay` is used.

Please notice that iOS Human Interface Guidelines say that the delay should be “a couple of seconds”.

This changes don’t affect the default behaviour of the class until we intentionally adjust `visibilityDelay`.

### DIFF
--- a/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.h
+++ b/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.h
@@ -57,6 +57,11 @@ NS_EXTENSION_UNAVAILABLE_IOS("Use view controller based solutions where appropri
 @property (readonly, nonatomic, assign) BOOL isNetworkActivityIndicatorVisible;
 
 /**
+ *  A time to wait before showing the network activity indicator. The default value is 0.
+ */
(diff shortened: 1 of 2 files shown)


lubiluk:
Yeah, my bad. Removed. I refactored the code a bit.

kcharwood:
This is interesting. I don't ever actually recall that part of the HIG, but its pretty clear:

>Display the network activity indicator to provide feedback when your app accesses the network for more than a couple of seconds. If the operation finishes sooner than that, you don’t have to show the network activity indicator, because the indicator is likely to disappear before users notice its presence.

I think this is worth pulling in, but I think there are some scenarios that still need to be thought through. I'll take a stab at adding to this in the morning.

Thanks for brining this to my attention!

:beers:

kcharwood:
I'm going to close this one out and direct conversation to #3094. Thanks for putting this together!

PR #3093, opened 2015-10-23
Use a valid version number for framework

tomtaylor:
iTunes Connect checks the framework versions when you upload your app binary. They need to be three non-negative integers, so you can't include `-beta.1` suffixes, for example.


kcharwood:
I'm going to close this one out until we are ready to release 3.0.0 properly.

PR #3094, opened 2015-10-23
A refactored version of #3086 allowing a delayed start to the network activity indicator

kcharwood:
This allows us to support the recommendation from the Apple HIG, which is short requests that last less than a couple of seconds should not trigger the network activity indicator.

Originally brought up in #3086 by @lubiluk 

cc @cnoon @kylef @jshier for review.

### DIFF
--- a/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
+++ b/UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
@@ -24,7 +24,15 @@
 #if TARGET_OS_IOS
 #import "AFURLSessionManager.h"
 
-static NSTimeInterval const kAFNetworkActivityIndicatorInvisibilityDelay = 0.17;
+typedef NS_ENUM(NSInteger, AFNetworkActivityManagerState) {
+    AFNetworkActivityManagerStateNotActive,
+    AFNetworkActivityManagerStateDelayingStart,
+    AFNetworkActivityManagerStateActive,
+    AFNetworkActivityManagerStateDelayingEnd
+};
+
+static NSTimeInterval const kDefaultAFNetworkActivityManagerActivationDelay = 1.0;
+static NSTimeInterval const kDefaultAFNetworkActivityManagerCompletionDelay = 0.17;
 
 static NSURLRequest * AFNetworkRequestFromNotification(NSNotification *notification) {
     if ([[notification object] respondsToSelector:@selector(originalRequest)]) {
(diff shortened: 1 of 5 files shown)


UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
```diff
@@ -148,6 +169,91 @@ - (void)networkRequestDidFinish:(NSNotification *)notification {
...
+                }
+                break;
+        }
+    }
+}
+
+- (void)startActivationDelayTimer {
+    self.activationDelayTimer = [NSTimer
+                                 timerWithTimeInterval:self.activationDelay target:self selector:@selector(activationDelayTimerFired) userInfo:nil repeats:NO];
+    [[NSRunLoop mainRunLoop] addTimer:self.activationDelayTimer forMode:NSRunLoopCommonModes];
```
cnoon: You probably want to add the timers to the `UITrackingRunLoopMode` mode as well for `iOS` targets. Otherwise you won't get called back when scrolling which is probably not what you want.

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
```diff
@@ -24,7 +24,15 @@
...
-static NSTimeInterval const kAFNetworkActivityIndicatorInvisibilityDelay = 0.17;
+typedef NS_ENUM(NSInteger, AFNetworkActivityManagerState) {
+    AFNetworkActivityManagerStateNotActive,
+    AFNetworkActivityManagerStateDelayingStart,
+    AFNetworkActivityManagerStateActive,
+    AFNetworkActivityManagerStateDelayingEnd
+};
+
+static NSTimeInterval const kDefaultAFNetworkActivityManagerActivationDelay = 1.0;
+static NSTimeInterval const kDefaultAFNetworkActivityManagerCompletionDelay = 0.17;
```
cnoon: Based on your original comments in the PR, should the activation delay be `2.0`? Also, where does the `0.17` come from? Seems pretty random to me.

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
```diff
@@ -75,28 +86,38 @@ - (instancetype)init {
...
     }
 }
 
-- (BOOL)isNetworkActivityIndicatorVisible {
+- (void)setNetworkingActivityActionWithBlock:(void (^)(BOOL networkActivityIndicatorVisible))block {
+    self.networkActivityActionBlock = block;
+}
+
+- (BOOL)isNetworkActivityOccurring {
     return self.activityCount > 0;
```
cnoon: I think there's a thread-safety issue here. The `activityCount` is being synchronized in all the places where the variable is being written to, but it is not being synchronized where the value is being read. I think this needs to be synchronized as well, otherwise I think we're racing to doom!

cnoon:
Overall I think this looks really good @kcharwood. Once you figure out what you'd like to do with the comments I pointed out, you've got a thumbs up from me. 👍🏼

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
kcharwood: That came from @mattt. Its been the value used in this class for all of `AFNetworking`.

Personally, two seconds seems like _forever_ to me. Thoughts on a sane default? It's configurable, so devs can change it to their preference.

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
kcharwood: [This](https://github.com/AFNetworking/AFNetworking/blob/master/UIKit%2BAFNetworking/AFNetworkActivityIndicatorManager.m#L112) is how AFN has been setup the whole time. I havent seen any issues caused by it in the past. Think I should still wrap it?

jshier: I think the only issue that could occur with non-synchronized reads is a ±1 inaccuracy. That is, someone reads the value while it's being incremented or decremented, but before the actual value is changed, leading the result to be ±1 from the actual value a moment later. This is probably a very tight race so it may not happen often, or at all. Whether it's worth the cost of `@synchronized` to avoid I couldn't say. For completeness, probably?

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
cnoon: I think a delay of 1 second seems sensible. I'm with you that 2 seconds seems like a really long time. I also think the 0.17 completion delay is fine. I was merely curious where that number came from.

UIKit+AFNetworking/AFNetworkActivityIndicatorManager.m
cnoon: I would just add a `@synchronized` around it just to be safe. Shouldn't really have much of any affect on performance in this case.

kcharwood:
Test failing due to unrelated issue. Will address those in another PR.

PR #3101, opened 2015-10-27
Custom collection field name formatter

pattrawoots:
I have been working with APIs that only take array parameters by repeating the name of variable without need of appending brackets '[]'. For example, http://test.com?test=value1&test=value2.

By applying these changes, developers will be able to customize the naming convention for collection parameters in query string.

### DIFF
--- a/AFNetworking/AFURLRequestSerialization.h
+++ b/AFNetworking/AFURLRequestSerialization.h
@@ -59,6 +59,9 @@ typedef NS_ENUM(NSUInteger, AFHTTPRequestQueryStringSerializationStyle) {
     AFHTTPRequestQueryStringDefaultStyle = 0,
 };
 
+typedef NSString * __nullable (^AFQueryStringNestedFieldNameFormatterBlock)(NSString * __nullable key, NSString *nestedKey);
+typedef NSString * __nullable (^AFQueryStringArrayFieldNameFormatterBlock)(NSString * key);
+
 @protocol AFMultipartFormData;
 
 /**
@@ -114,6 +117,16 @@ typedef NS_ENUM(NSUInteger, AFHTTPRequestQueryStringSerializationStyle) {
  @see NSMutableURLRequest -setTimeoutInterval:
  */
 @property (nonatomic, assign) NSTimeInterval timeoutInterval;
+
+/**
+ The field name formatter for nested field in query string.
+ */
+@property (nonatomic, copy) AFQueryStringNestedFieldNameFormatterBlock nestedFieldNameFormatterBlock;
+
+/**
+ The field name formatter for array field in query string.
+ */
+@property (nonatomic, copy) AFQueryStringArrayFieldNameFormatterBlock arrayFieldNameFormatterBlock;
 
 ///---------------------------------------
 /// @name Configuring HTTP Request Headers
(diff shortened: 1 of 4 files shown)


kcharwood:
Thanks for the PR! This is interesting. I personally haven't ran into this issue before. We expose `setQueryStringSerializationWithBlock` on `AFURLRequestSerialization` to allow anyone to do their own custom query string serialization, so this comes down to balancing complicating the API interface vs making it easier for "corner case" development.

We're there any particular shortcomings with `setQueryStringSerializationWithBlock` that made you go down this route?

@cnoon @jshier @kylef Any feelings one way or the other on if this belongs in core?

kcharwood:
Bump for @cnoon @jshier @kylef Any feelings one way or the other on if this belongs in core?

cnoon:
I'm with you @kcharwood...this is certainly an interesting approach. At the moment, I lean towards this being a bit too complicated. It just clutters the APIs IMO. I do think there is some nice functionality here though for anyone that needs to customize this type of behavior.

I'm not in favor of merging this PR as is. I would really REALLY like to @pattrawoots reply to:

> We're there any particular shortcomings with setQueryStringSerializationWithBlock that made you go down this route?

If you don't get a reply, then I'd probably just close this for now and add a card to Trello to investigate simplifying this PR and intermixing with the `setQueryStringSerializationWithBlock` approach.

pattrawoots:
Hi @cnoon and @kcharwood,

I fully understand that I'd better go with setQueryStringSerializationWithBlock to customize the serialization.

However, it would be great if some helpful AF's functions were made public. (e.g. AFPercentEscapedStringFromString), as I just don't want to write a whole new serializer that is likely to look the same as AFQueryStringFromParameters and AFQueryStringPairsFromKeyAndValue with a few differences.

What do you think?

kcharwood:
Exposing those utilities could certainly be valuable. Feel free to submit a PR with them exposed and tests added!

:beers:

pattrawoots:
I just found that there is no way to customize parameter name when sending a multipart form request which is the case I'm experiencing. This is because setQueryStringSerializationWithBlock is only applied to non-multipart form one.

As you can see in multipartFormRequestWithMethod:, it only calls AFQueryStringPairsFromDictionary to construct a body.

Anyway, I'm working on a new PR to get rid of this limitation. :)

cnoon:
Sounds good @pattrawoots. In Alamofire we've already exposed these methods for exactly this reason.

pattrawoots:
hi @cnoon, will the method expose also enable developers to control the name of collection parameter when they use multipartFormRequestWithMethod:? 

As mentioned in my latest comment, multipartFormRequestWithMethod: internally calls AFQueryStringPairsFromDictionary, and developers cannot change the behavior of this method as it doesn't not have any property like setQueryStringSerializationWithBlock.

I would be appreciate it if you could share what the code would look like after the methods get exposed. :)

cnoon:
We did expose that method, but not the ability to modify the naming convention. We also exposed the parameter escaping method which is the important one. Here's a [link](https://github.com/Alamofire/Alamofire/blob/master/Source/ParameterEncoding.swift#L176-L250) to the Alamofire logic that handles this same type of functionality.

PR #3117, opened 2015-11-04
Fixed race condition in reachability callback delivery

MichaelHackett:
This is an alternative implementation of a previous attempt to fix a race condition in reachability notifications, particularly with domain-name-based associations. However, we encountered cases where the status-change callback could still receive notifications out of order, due to one path executing the callback immediately on the runloop thread and the other queuing the callback onto the GCD main queue (which is _serviced_ by the same run loop, but the order in which queue blocks and other run loop callbacks are executed is not easy to determine).

I broke this into two commits: The first is the fix for the issue, by sending all callback notifications through the GCD main queue, just as the NSNotifications were being sent. This is the code we've been using for a few weeks and it has resolved all our issues. The second removes the special case for domain-based reachability associations, as I don't think there's any need for it with this implementation. However, I have not tested this; I just noticed that there was an attempt to remove it a few months ago, but it was quickly restored due to reintroducing the race condition. If I've really solved the root cause, then it should be safe to remove this now. But I can't think of a way to unit-test this, so it needs to be tried in a few different apps to see if it's safe. If not, I can reverse that commit and resubmit with the first alone.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -76,20 +76,28 @@ static AFNetworkReachabilityStatus AFNetworkReachabilityStatusForFlags(SCNetwork
     return status;
 }
 
-static void AFNetworkReachabilityCallback(SCNetworkReachabilityRef __unused target, SCNetworkReachabilityFlags flags, void *info) {
+/**
+ * Queue a status change notification for the main thread.
+ *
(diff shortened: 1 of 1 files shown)


kcharwood:
This looks great! Thanks for putting this together. I should be able to get this merged in today, and I plan on shipping it in the 2.6.2 release.

MichaelHackett:
I appreciate the comment! I'm glad it's of use.

Issue #2810, opened 2015-06-25
Swift 2.0 compatibility issues


johansson:
I understand Swift 2.0 is still beta and all, but with the new `throws` behavior in light, I'd like Swift 2.0 users to be able to use AFNetworking without `throws` throwing it off.

The basis is simple, if you used the overridden serializer:

```swift
class FooSerializer: AFHTTPResponseSerializer {
    override func responseObjectForResponse(response: NSURLResponse!, data: NSData!) throws -> AnyObject {
        let response = try super.responseObjectForResponse(response, data: data)
        return FooResponse(data: response)
    }
}
```

and you get a HTTP error (401, for example). `AFHTTPRequestOperation` does not call the `failure` block, and instead calls the `success` block. However, if you don't use the subclassed `AFHTTPResponseSerializer` and simply use `AFHTTPResponseSerializer`, the `AFHTTPRequestOperation` correctly calls the `failure` block.

You can see the full code of my minimum test case for this bug [here](https://github.com/johansson/AFNetworkingTest). But I'll copy the important code here for context:

```swift
let s = FooSerializer()

let r_200 = NSURLRequest(URL: NSURL(string: "http://localhost:8080/200")!)
let r_401 = NSURLRequest(URL: NSURL(string: "http://localhost:8080/401")!)

let o1 = AFHTTPRequestOperation(request: r_200)

o1.responseSerializer = s
o1.setCompletionBlockWithSuccess({
    (operation, result) -> Void in
        NSLog("o1 success - EXPECTED")
}) {
    (request, error) -> Void in
        NSLog("o1 failure - NOT EXPECTED")
}

let o2 = AFHTTPRequestOperation(request: r_401)
o2.responseSerializer = s
o2.setCompletionBlockWithSuccess({
    (operation, result) -> Void in
        NSLog("o2 success - NOT EXPECTED")
}) {
    (request, error) -> Void in
        NSLog("o2 failure - EXPECTED")
}

let o3 = AFHTTPRequestOperation(request: r_401)
o3.responseSerializer = AFHTTPResponseSerializer()
o3.setCompletionBlockWithSuccess({
    (operation, result) -> Void in
        NSLog("o3 success - NOT EXPECTED")
}) {
    (request, error) -> Void in
        NSLog("o3 failure - EXPECTED")
}

o1.start()
o2.start()
o3.start()
```

The output of the above is something like:

```
[...]

PR #3130, opened 2015-11-06
Fixed swift interop issue with throws and Request/Response serialization

kcharwood:
Fix for #2810


kcharwood:
Merged into 3_0_0 in ad83475

Issue #3110, opened 2015-11-03
AFNetworkReachabilityManager.m内存泄漏


Ruoyiming:
```
AFNetworking/AFNetworkReachabilityManager.m:130:45: Call to function 'SCNetworkReachabilityCreateWithName' returns a Core Foundation object with a +1 retain 
AFNetworking/AFNetworkReachabilityManager.m:132:5: Object leaked: object allocated and stored into 'reachability' is not referenced later in this execution path and has a retain count of AFNetworking/AFNetworkReachabilityManager.m:139:45: Call to function 'SCNetworkReachabilityCreateWithAddress' returns a Core Foundation object with a +1 retain AFNetworking/AFNetworkReachabilityManager.m:141:5: Object leaked: object allocated and stored into 'reachability' is not referenced later in this execution path and has a retain count of +1
```

kcharwood:
Can you let me know how you generated those warnings? I'm not seeing them locally. Thanks

rsilva:
@kcharwood I'm seeing this issue as well. 

1. Create new iOS project using Xcode 7.1. (single view, objective-c)
2. Add AFNetworking files 
3. Open project settings and set `Analyze During 'Build'` to `YES`

<img width="1426" alt="screen shot 2015-11-03 at 10 21 35 am" src="https://cloud.githubusercontent.com/assets/1325508/10912298/d50e40e8-8216-11e5-9850-ba3fdc9bdb17.png">

Ruoyiming:
Yes. In Xcode  7.1, when I run the project, It also generates a warning of NSURLConnection.

kcharwood:
These aren't actually leaked, as you can see `CFBridgingRelease` a few lines below. Let me see if I can suppress them.

kcharwood:
Thanks!

Ruoyiming:
@kcharwood if you want ignore error, you can change it like bellow
#ifndef __clang_analyzer__
+ (instancetype)managerForDomain:(NSString *)domain {

    SCNetworkReachabilityRef reachability = SCNetworkReachabilityCreateWithName(kCFAllocatorDefault, [domain UTF8String]);

    AFNetworkReachabilityManager *manager = [[self alloc] initWithReachability:reachability];
    return manager;
}
#endif

#ifndef __clang_analyzer__
+ (instancetype)managerForAddress:(const void *)address {
[...]

PR #3131, opened 2015-11-06
Suppressed false positive memory leak warning in Reachability Manager

kcharwood:
Fix for #3110


kcharwood:
Merged into 3.0.0 with 7d8fba1

Issue #3104, opened 2015-10-28
NSURL return value from setDownloadTaskDidFinishDownloadingBlock should be optional / nullable


alfiehanssen:
The method is declared like so in AF v2.6.1: 

```
- (void)setDownloadTaskDidFinishDownloadingBlock:(nullable NSURL * (^)(NSURLSession *session, NSURLSessionDownloadTask *downloadTask, NSURL *location))block;
```

This block is called like so: 

```
#pragma mark - NSURLSessionDownloadDelegate

- (void)URLSession:(NSURLSession *)session
      downloadTask:(NSURLSessionDownloadTask *)downloadTask
didFinishDownloadingToURL:(NSURL *)location
{
    AFURLSessionManagerTaskDelegate *delegate = [self delegateForTask:downloadTask];
    if (self.downloadTaskDidFinishDownloading) {
        NSURL *fileURL = self.downloadTaskDidFinishDownloading(session, downloadTask, location);
        if (fileURL) {
            delegate.downloadFileURL = fileURL;
            NSError *error = nil;
            [[NSFileManager defaultManager] moveItemAtURL:location toURL:fileURL error:&error];
            if (error) {
                [[NSNotificationCenter defaultCenter] postNotificationName:AFURLSessionDownloadTaskDidFailToMoveFileNotification object:downloadTask userInfo:error.userInfo];
            }

            return;
        }
    }

    if (delegate) {
        [delegate URLSession:session downloadTask:downloadTask didFinishDownloadingToURL:location];
    }
}
```

The invocation of the block indicates that the NSURL return value can be nil. And there are cases where we want it to be nil. But consuming this API in Swift the return value is not optional. Forcing us to return a non-nil value i.e. forcing us to invent a bogus value.

kcharwood:
dupe of #3043

Issue #3043, opened 2015-10-13
Invalid nullability attributes for AFURLSessionManager


dimazen:
AFURLSessionManager in 2.6 contains following method:

```
- (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request
                            completionHandler:(nullable void (^)(NSURLResponse *response, id responseObject, NSError *error))completionHandler;
```
Taking into account `NS_ASSUME_NONNULL_BEGIN` at the beginning of the file it means, that both `responseObject` and `error` won't be nil at the same time. 
Can be fixed by:
```
(nullable void (^)(NSURLResponse *response, nullable id responseObject, nullable NSError *error))
```

Also: 
In 2.5 AFURLSessionManager contains callback for task completion which is imported into Swift as:
```
public func setDownloadTaskDidFinishDownloadingBlock(block: ((NSURLSession!, NSURLSessionDownloadTask!, NSURL!) -> NSURL!)!)
```

As you see, NSURL is optional here, also implementation handles this NSURL as an optional value as well:

```
NSURL *fileURL = self.downloadTaskDidFinishDownloading(session, downloadTask, location);
if (fileURL) {
}
```

However in 2.6 this callback declared as:
```
public func setDownloadTaskDidFinishDownloadingBlock(block: ((NSURLSession, NSURLSessionDownloadTask, NSURL) -> NSURL)?)
```
But implementation is the same (i.e. allows optional values). 
Returning `nil`, as far as I know, instruct system to remove that file automatically, therefore if for some reason I can not return URL there is no reason for me to worry about that file.

kcharwood:
Hi @dimazen,

Your first issue is already addressed in the latest master branch.

As for the second, I'm not sure that should be an optional NSURL returned. Can you explain the use case you see for that?

sowenjub:
Hi @kcharwood, I'm facing the same issue with the following use case: I use setDownloadTaskDidFinishDownloadingBlock to save the temp file into CoreData, so I don't need the temp file to be moved anywhere.

dimazen:
@kcharwood 
Regarding second issue: 

My code looks like that: 

```
tasksManager.setDownloadTaskDidFinishDownloadingBlock { [weak self] session, task, url in
    return self?.downloadURLForTask(task)
}
```
[...]

PR #3133, opened 2015-11-06
Fixed swift interop issue that prevented returning a nil NSURL for a download task

kcharwood:
Fix for #3104 and #3043


kcharwood:
Brought into 3.0.0 with e501c8e

Issue #3119, opened 2015-11-05
Adding a CA does not automatically make its certificates pass with AFSecurityPolicy


rafaelc0sta:
Adding a custom CA to AFSecurityPolicy does not seem to make its child certificates pass through -evaluateServerTrust... This is (or at least, should be) unexpected behaviour, as adding a CA makes much more sense than adding the whole tree (CA and certificate). Creating a "AFSSLPinningModeCA" pinning mode would solve the problem, and should be an easy fix.

Update: Updating the code to work as I expect it to, I saw that using allowInvalidCertificates produces an equal input (at first glance, don't know for sure). It's important to note that an invalid/expired certificate is very different than a self-signed non-expired one. Here's my modification though... I can submit a pull request any time :) https://github.com/AFNetworking/AFNetworking/compare/master...rafaelc0sta:master

kcharwood:
Would you be able to submit a PR with tests?

rafaelc0sta:
@kcharwood Will submit! Sidenote: I wasn't able to run the tests on my machine (not that they failed, I just couldn't run them).

kcharwood:
Closing this. Reference #3159

rafaelc0sta:
@kcharwood Thanks! Look forward to see #3159 on 3.0.0! :)

kcharwood:
I released beta 2 earlier today, which has it in there. Thanks again!

PR #2921, opened 2015-08-25
Fix certificate validation for servers providing incomplete chains

p4checo:
Some servers send incomplete certificate chains (without the Root CA), which cause issues in the latest version's (2.6.0) certificate pinning mechanism. In these cases, when `AFSecurityPolicy`s `-evaluateServerTrust:` invokes `AFCertificateTrustChainForServerTrust(serverTrust)` (prior to setting the
anchor), the Root CA isn't present (the server didn't send it).

Afterwards, even though `AFServerTrustIsValid(serverTrust)` succeeds, the next validation fails since no pinned certificate matches any of the `serverCertificates`.

By fetching the `AFCertificateTrustChainForServerTrust(serverTrust)` *after* the `AFServerTrustIsValid(serverTrust)` validation, the complete chain is obtained and the Root CAs match (pinned vs validated).

A slight optimization was also made when validating if the `serverTrust` certificates contain the `pinnedCertificates`, looping in reverse order and returning after finding the first match.

I've included a new test that demonstrates the mentioned issue is fixed by this patch.

PR #3136, opened 2015-11-06
Added AFSSLPinningModeCA and tests

rafaelc0sta:
Added AFSSLPinningModeCA and tests

### DIFF
--- a/Tests/Tests/AFSecurityPolicyTests.m
+++ b/Tests/Tests/AFSecurityPolicyTests.m
@@ -68,6 +68,24 @@ static SecTrustRef AFUTHTTPBinOrgServerTrustLeafOnly() {
 	return trust;
 }
 
+static SecTrustRef AFUTADNNetServerTrustLeafOnly() {
+	NSString *certPath = [[NSBundle bundleForClass:[AFSecurityPolicyTests class]] pathForResource:@"ADNNetServerTrustChain/adn_0" ofType:@"cer"];
+	NSCAssert(certPath != nil, @"Path for certificate should not be nil");
+	NSData *certData = [NSData dataWithContentsOfFile:certPath];
+	
+	
+	
+	NSMutableArray *certs  = [NSMutableArray arrayWithCapacity:1];
+	SecCertificateRef cert = SecCertificateCreateWithData(NULL, (__bridge CFDataRef)(certData));
+	[certs addObject:(__bridge id)(cert)];
+	
+	
+	SecPolicyRef policy = SecPolicyCreateBasicX509();
+	SecTrustRef trust = NULL;
+	SecTrustCreateWithCertificates((__bridge CFTypeRef)(certs), policy, &trust);
+	return trust;
+}
+
 static SecCertificateRef AFUTHTTPBinOrgServerTrustAboveLeaf() {
 	NSString *certPath = [[NSBundle bundleForClass:[AFSecurityPolicyTests class]] pathForResource:@"HTTPBinOrgServerTrustChain/httpbin_1" ofType:@"cer"];
 	NSCAssert(certPath != nil, @"Path for certificate should not be nil");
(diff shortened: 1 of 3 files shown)


kcharwood:
Linked to #3119

Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -347,6 +374,14 @@ - (void)testPolicyWithCertificatePinningAllowsHTTPBinOrgServerTrustWithHTTPBinOr
     XCTAssertTrue([policy evaluateServerTrust:AFUTHTTPBinOrgServerTrust() forDomain:nil], @"Policy should allow server trust");
 }
 
+- (void)testPolicyWithCertificatePinningAllowsHTTPBinOrgServerTrustWithCertificateAbovePinned {
```
kcharwood: Thanks for adding a test. Can you add some negative tests? When it comes to security, its more important to verify that things don't work when they should not, since that leads to vulnerabilities.

rafaelc0sta:
@kcharwood Added the negative tests!

kcharwood:
Thanks @rafaelc0sta 

I'm always _super_ cautious with changes to `AFSecurityPolicy`, so I like to make sure they are all properly vetted before pulling them in.

Can you please explain in clear numbered steps what needs to happen for a certificate authority to be considered valid here? I'd like to read through that while cross referencing your patch here.

Thanks again for taking the time on this!

kcharwood:
@reedloden I would love your thoughts on this one.

rafaelc0sta:
@kcharwood Sure! By adding this mode, one makes any certificate signed by a CA (the one which is pinned) valid. This way, you can pin the CA and any certificates below it will be valid (a behaviour seen in Retrofit, for example). This change adds the mode separately. One would think that the Certificate Pinning mode would behave as this mode I added, but it seems like it validates the whole chain. Anyway: the code which runs when CA is selected adds the CA (SecTrustSetAnchorCertificates) and then evaluates the server's certificate to check whether it was signed by the CA or not. If it was, it returns true. If it wasn't, false. I even used AFNetworking's own method to return.

That's it.

p4checo:
Hi guys,

I really don't think it's necessary to create another `AFSSLPinningMode`. 
By looking at the code, the logic in the new `AFSSLPinningModeCA` is essentially the same as `AFSSLPinningModeCertificate`, except for the validation that at least one of the certificates in the received `serverTrust` chain is in fact contained in the `self.pinnedCertificates`.

What `SecTrustSetAnchorCertificates` does is define "authoritative entities for which trust is assumed and not derived" (https://en.wikipedia.org/wiki/Trust_anchor). So any certificate added as anchor is automatically trusted and validates any certificate chain below it (if properly signed).

In fact, defining the anchors as *only* being the pinned certificates should probably be enough to provide the necessary protection against MiTM attacks, if the **Root CA's** are the pinned certificates. I guess the extra validation to cross check the contents of the `pinnedCertificates` with the `serverTrust` chain was added as a precaution measure (which created this corner case :smile:).

I might be mistaken, but it seems to me that @rafaelc0sta was having the same issue with incomplete chains as I did (#2921). The server didn't provide the complete chain (most likely didn't send the Root CA), and even though the `AFServerTrustIsValid()` returned `YES` (the anchor validated), the pinned certificate wasn't in the `serverCertificates` array (which should be filled *after* validating the trust, to have the complete chain).

@rafaelc0sta: were you using any public API, and have you tested with my branch and see if your problem persists?

Cheers!

rafaelc0sta:
@p4checo I agree with you Andre, this behaviour should be the one in AFSSLPinningModeCertificate, and yes, the AFServerTrustIsValid() does return TRUE. I only added another mode because I thought that the behaviour seen in AFSSLPinningModeCertificate was a design choice. But as you said, with the current implementation, there is a corner case. I believe your problem and mine are the same, and if I had to choose, fixing the certificate pinning mode would be ideal (#2921)

p4checo:
@rafaelc0sta: is the API you're using (which exhibits this behavior) public?

kcharwood:
Thanks for the write up @p4checo. If you know of an endpoint I could test against, that would be _very_ helpful

rafaelc0sta:
@p4checo My implementation only makes the code check for the validity of the certificate with the pinned CA. It only makes the AFServerTrustIsValid be returned instead of not being called at all because of the chain.

p4checo:
@kcharwood: I'll try and find one. :smile: However, IMO I don't think that should be a requirement for the MR to be merged. I've added a test case that illustrates the issue and passes on my branch but fails in 2.6.0. All the other tests pass too. Still, I understand the extra caution here. :stuck_out_tongue:

kcharwood:
Ya obviously I would feel much more confident stepping through a real implementation, just to give me a better understanding of the use case. Definitely don't want a regression here :)

kcharwood:
Thanks again @rafaelc0sta for bringing this up. At this point, I think I would prefer to not add another value to the enum, and stick with something more in line with @p4checo's approach.

If you are cool with that, I'm going to close this one out. Thanks for the discussion here!

kcharwood:
See #3159

PR #3148, opened 2015-11-11
Migrated to Fastlane for CI and Deployment

kcharwood:
Documenting the shared fastlane support [here](https://github.com/AFNetworking/fastlane), which is being designed to use across multiple open source projects.


.gitignore
```diff
@@ -19,3 +19,20 @@ DerivedData
...
+/fastlane/Error*.png
+# deliver temporary preview output
+/fastlane/Preview.html
+# snapshot generated screenshots
+/fastlane/screenshots/*/*-portrait.png
+/fastlane/screenshots/*/*-landscape.png
+/fastlane/screenshots/screenshots.html
+# frameit generated screenshots
+/fastlane/screenshots/*/*-portrait_framed.png
+/fastlane/screenshots/*/*-landscape_framed.png
```
kylef: Can the above 7 lines just be simplified to `fastlane/screenshots`?

kcharwood: Ya. This is the copy-paste from the [fastlane docs](https://github.com/fastlane/fastlane/blob/master/docs/Advanced.md#gitignore)

kcharwood: We're not even using snapshot, so it could be pulled altogether

kcharwood:
The indicator view tests continue to cause a stalling problem randomly on travis, so I'm commenting those out for the time being.

PR #3150, opened 2015-11-12
Fixed AFImageDownloader stalling after numerous failures

rsilva:
The active task count was only being decremented after success. Once the number of failures reaches the maximumActiveDownloads, no more tasks were being started.


kcharwood:
Thank you! I bet this was contributing to some testing issues I was running into.

PR #2502, opened 2015-01-06
fix bug for AFMultipartBodyStream

bang590:
may cause inputStream read more bytes than the setting value numberOfBytesInPacket

PR #3153, opened 2015-11-12
Fixed bug in AFMultipartBodyStream that may cause the input stream to read more bytes than required.

kcharwood:
may cause inputStream read more bytes than the setting value numberOfBytesInPacket


kcharwood:
From #2502

PR #3158, opened 2015-11-13
add POST api with progress

dake:
add new POST api to support progress

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -181,10 +181,10 @@ didCompleteWithError:(NSError *)error
     } else {
         dispatch_async(url_session_manager_processing_queue(), ^{
             NSError *serializationError = nil;
-            responseObject = [manager.responseSerializer responseObjectForResponse:task.response data:data error:&serializationError];
-
             if (self.downloadFileURL) {
                 responseObject = self.downloadFileURL;
+            } else {
+                responseObject = [manager.responseSerializer responseObjectForResponse:task.response data:data error:&serializationError];
             }
 
             if (responseObject) {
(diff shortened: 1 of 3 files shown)


kcharwood:
Thanks for putting this together? Would you be willing to add a test demonstrating the progress is called as expected? Perhaps point it to the post action on  `http://httpbin.org/`.

Thanks!

dake:
hi, I need post api a progress parameter when upload a file stream with post form block. so I added a progress parameter in post form api.

AFNetworking/AFURLSessionManager.m
```diff
@@ -181,10 +181,10 @@ - (void)URLSession:(__unused NSURLSession *)session
     } else {
         dispatch_async(url_session_manager_processing_queue(), ^{
             NSError *serializationError = nil;
-            responseObject = [manager.responseSerializer responseObjectForResponse:task.response data:data error:&serializationError];
-
             if (self.downloadFileURL) {
                 responseObject = self.downloadFileURL;
+            } else {
+                responseObject = [manager.responseSerializer responseObjectForResponse:task.response data:data error:&serializationError];
```
kcharwood: Does this change have anything to do with this PR?

kcharwood:
I wonder if this opens up the API to more one off methods? Currently this allows you to easily get NSProgress for a multipart POST, but would someone also want it for a regular POST or PUT?

Can you use the `progressForTask:` method and use the returned task to grab your progress? At the moment, I'm leaning towards keeping the API the same just to make sure we don't over complicate it with lots of options.

AFNetworking/AFURLSessionManager.m
dake: No, this is another fix, causing performance

kcharwood:
Take a look at #3187

PR #3159, opened 2015-11-13
Fixed certificate validation for servers providing incomplete chains

p4checo:
Some servers send incomplete certificate chains (without the Root CA), which
cause issues in the latest version's (2.6.3) certificate pinning mechanism.
In these cases, when `AFSecurityPolicy`s `-evaluateServerTrust:` invokes
`AFCertificateTrustChainForServerTrust(serverTrust)` (prior to setting the
anchor), the Root CA isn't present (the server didn't send it). Afterwards,
even though `AFServerTrustIsValid(serverTrust)` succeeds, the next validation
fails since no pinned certificate matches any of the `serverCertificates`.

By fetching the `AFCertificateTrustChainForServerTrust(serverTrust)` *after*
the `AFServerTrustIsValid(serverTrust)` validation, the complete chain is
obtained and the Root CAs match (pinned vs validated).

A slight optimization was also made when validating if the `serverTrust`
certificates contain the `pinnedCertificates`, looping in reverse order and
returning after finding the first match.

Follow up of PR #2921, but rebased for 3.0.0


Tests/Tests/AFSecurityPolicyTests.m
```diff
@@ -388,6 +418,28 @@ - (void)testPolicyWithCertificatePinningAllowsHTTPBinOrgServerTrustWithHTTPBinOr
     XCTAssertTrue([policy evaluateServerTrust:AFUTHTTPBinOrgServerTrust() forDomain:@"httpbin.org"], @"Policy should allow server trust");
 }
 
+- (void)testPolicyWithCertificatePinningAllowsGoogleComServerTrustIncompleteChainWithRootCertificatePinnedAndValidDomainName {
+    // google.com sends an incomplete certificate chain consisting of 3 certificates ( *.google.com, Google Internet Authority G2 and GeoTrust Global CA)
```
kcharwood: Could you update the comment with a reference to the pull request number? Thanks for the docs here.

p4checo: Sure

kcharwood:
Are there any negative tests we should consider here?

p4checo:
I think the negative tests are the same as the "classic" certificate pinning case, i.e. if none of the server´s certificates is pinned, the connection should be cancelled. The same goes for expired and invalid certificates, I guess.

I can't think of a specific negative case for this particular issue. Any idea?

kcharwood:
I couldn't think of any either, just wanted to pose the question.

p4checo:
:tada: :beers: :tada:

kcharwood:
@p4checo it looks like expiration day has come so the test suite is now failing.

Can you walk me through where you grabbed these two chains so I can get them updated?

p4checo:
Hi @kcharwood,

I used https://www.ssllabs.com/ssltest/analyze.html?d=google.com to validate whether it was using two chains, and then I got the certificates using OpenSSL:

`openssl s_client -showcerts -connect google.com:443 </dev/null 2>/dev/null`
(I think I ran the command multiple times for the both chains to appear, but I'm not sure)

However, I've now checked that google.com seems to no longer have 2 certification paths :sweat:

kcharwood:
How should we proceed here? I don't want to disable the test if I don't have too.

p4checo:
Ideally, we should find another case of a multiple certification path host and update the test. :smile:

At the time I was lucky google.com itself was an example. But surely there are other cases.

kcharwood:
K. I don't have a lot of time to dig on this issue at the moment, so I may need to disable it in the near term to allow the test suite to pass for some other issues I need to push through.

p4checo:
Me neither :disappointed:. If I manage to put aside some time for this I'll let you know.

RadimC:
If we've set our pinned certs as the exclusive anchors in our trust store, why do we need to explicitly compare cert data at all?

```
NSMutableArray *pinnedCertificates = [NSMutableArray array];
for (NSData *certificateData in self.pinnedCertificates) {
    [pinnedCertificates addObject:(__bridge_transfer id)SecCertificateCreateWithData(NULL, (__bridge CFDataRef)certificateData)];
}
SecTrustSetAnchorCertificates(serverTrust, (__bridge CFArrayRef)pinnedCertificates);
 
return AFServerTrustIsValid(serverTrust);
```

What problem is the additional code solving?

p4checo:
The issue this PR was fixing was due to incomplete certificate chains being sent by the server and making certificate pinning fail precisely because of this certificate data comparison.

However, the point you raise is valid since the certificates are already defined as anchors, we shouldn't have to compare the data ourselves manually. That should be (and already is) done more efficiently by `SecTrustEvaluate` (via `AFServerTrustIsValid`), right?

So, that data comparison check (and my complete "fix") would be unnecessary because `AFServerTrustIsValid` would handle it automatically by using the defined anchors. I guess I didn't see the complete picture back then 😅 

I heard they welcome PR's in this project, @RadimC 😉 

Cheers! 🍻

PR #3164, opened 2015-11-18
Changed `pinnedCertificates` type in `AFSecurityPolicy` from `NSArray` to `NSSet`

0xced:
Pinned certificates were deduplicated anyway, so let’s make it clear directly in the API.


0xced:
Although this should not impact v2 to v3 migration, this should probably be mentioned somewhere. What’s the best place, the CHANGELOG or the wiki?

kcharwood:
Now seems like a great time to do this. 

Both the Changelog and the wiki would be a good place to include this.

kcharwood:
The issue will automatically get created in the Changelog, but I still need to update the wiki.

Thanks again!

:beers:

PR #3166, opened 2015-11-19
Added support for Objective-C light weight generics

kcharwood:
Since v3 is going to require Xcode 7, adding generics support would be nice. Who knows, it might even catch some lingering bugs. 😝


kcharwood:
Added a PR here to this issue @0xced

AFNetworking/AFSecurityPolicy.h
```diff
@@ -71,7 +71,7 @@ NS_ASSUME_NONNULL_BEGIN
 
  @return The default security policy.
```
kcharwood: I noticed this documentation was wrong. Fixed it in the 3_0_0 branch

PR #3169, opened 2015-11-19
Better error when the server trust is invalid

### LINKED ISSUE #3165: Can’t distinguish between server trust evaluation failure and task cancellation
When server trust evaluation fails, AFNetworking calls the completion handler with disposition `NSURLSessionAuthChallengeCancelAuthenticationChallenge` which in turn calls NSURLSessionTask’s `cancel` method.

The resulting error is from `NSURLErrorDomain` with code `NSURLErrorCancelled` (-999) which is the exact same error you get if you call `cancel` on a task yourself.

I think it’s AFNetworking’s responsibility to properly handle server trust failure and report a better error.

I think we should introduce
```objc
- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
                  forDomain:(NSString *)domain
                      error:(NSError **)error
```
, construct a better error and make it bubble up to the `AFHTTPSessionManager` failure handlers.

0xced:
When the server trust is invalid, using `NSURLSessionAuthChallengeCancelAuthenticationChallenge` terminates the task with an error from `NSURLErrorDomain` with code `NSURLErrorCancelled` (-999) which is indistinguishable from the error you get when calling the `cancel` method on a `NSURLSessionTask`.

Using `NSURLSessionAuthChallengeRejectProtectionSpace` instead produces a much better error:

```
Error Domain:           NSURLErrorDomain
Code:                   NSURLErrorServerCertificateUntrusted (-1202)
NSLocalizedDescription: "The certificate for this server is invalid. You might be connecting to a server that is pretending to be “httpbin.org” which could put your confidential information at risk."
```

Fixes #3165

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,10 @@ script:
   - set -o pipefail
   - fastlane $FASTLANE_LANE configuration:Debug --env $FASTLANE_ENV
   - fastlane $FASTLANE_LANE configuration:Release --env $FASTLANE_ENV
+after_failure:
+  - cat -n ~/Library/Logs/scan/*
+  - cat -n $TMPDIR/com.apple.dt.XCTest-status/Session*.log
+  - cat -n ~/Library/Logs/DiagnosticReports/xctest*.crash
 # Commenting Deploy out until out of beta
 # deploy:
 #   provider: script
--- a/AFNetworking/AFSecurityPolicy.m
+++ b/AFNetworking/AFSecurityPolicy.m
@@ -94,6 +94,7 @@ static BOOL AFServerTrustIsValid(SecTrustRef serverTrust) {
     SecTrustResultType result;
     __Require_noErr_Quiet(SecTrustEvaluate(serverTrust, &result), _out);
 
+    NSLog(@"✳️ %s %d ✳️ SecTrustEvaluate -> %d", __PRETTY_FUNCTION__, __LINE__, result);
     isValid = (result == kSecTrustResultUnspecified || result == kSecTrustResultProceed);
 
 _out:
@@ -269,9 +270,11 @@ static NSArray * AFPublicKeyTrustChainForServerTrust(SecTrustRef serverTrust) {
             for (NSData *certificateData in self.pinnedCertificates) {
                 [pinnedCertificates addObject:(__bridge_transfer id)SecCertificateCreateWithData(NULL, (__bridge CFDataRef)certificateData)];
             }
+			NSLog(@"✳️ %s %d ✳️ pinnedCertificates = %@", __PRETTY_FUNCTION__, __LINE__, pinnedCertificates);
             SecTrustSetAnchorCertificates(serverTrust, (__bridge CFArrayRef)pinnedCertificates);
 
             if (!AFServerTrustIsValid(serverTrust)) {
+				NSLog(@"✳️ %s %d ✳️ !AFServerTrustIsValid", __PRETTY_FUNCTION__, __LINE__);
                 return NO;
             }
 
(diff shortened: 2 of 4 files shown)


0xced:
As discussed in #3165, here is a PR to see what happens on travis.

The error in unit test is NSURLErrorServerCertificateUntrusted (-1202) but the error with the same code in the demo app is NSURLErrorSecureConnectionFailed (-1200).

0xced:
The test passes both on the iOS 8.4 simulator and on OS X on my machine with El Capitan 10.11.1 and Xcode 7.1.1.

kcharwood:
On Travis, the test passed on iOS 9.0, 9.1, and tvOS 9.0? I wonder what could be the difference between your setup?

I'm away from my computer this weekend, so may not be able to dig any further until Monday.

cnoon:
Not positive, but I'm guessing the differences here in the behavior is caused by ATS.

0xced:
I thought about ATS too but I don’t see how it could explain why the test is failing on iOS 8 and passing on iOS 9.

Also, I can’t reproduce on my machine so I’m thinking about using [Xtrace][1] to understand the exact code path.

  [1]: https://github.com/johnno1962/Xtrace

cnoon:
That definitely sounds like a good approach @0xced.

0xced:
OK, so I can reproduce the failure in the iOS 8.4 simulator when `testInvalidServerTrustProducesCorrectError` is run as part of the AFHTTPSessionManager test case. If I run the test individually, it passes!

So I’m suspecting that calling `[self.manager invalidateSessionCancelingTasks:YES];` in the `AFHTTPSessionManagerTests` tear down is causing some unexpected side effects. I will investigate.

0xced:
I deleted the `tearDown` method but `testInvalidServerTrustProducesCorrectError` is still failing when run as part of the full test case.

But at least I have a reproducible way to make it pass and a reproducible way to make it fail so understanding this issue is now only a matter of time.

kcharwood:
Changing the test to hit an HTTPS endpoint that is _only_ hit by this test seems to work for me on iOS 8.1.

 As discussed on Slack, I'm assuming this is because the TLS cache is outside of our control, so iOS is reusing an existing valid TLS session for that endpoint, and not consulting the challenge block.

Something like this should work:

```
- (void)testInvalidServerTrustProducesCorrectError {
    __weak XCTestExpectation *expectation = [self expectationWithDescription:@"Request should fail"];
    NSURL *googleCertificateURL = [[NSBundle bundleForClass:[self class]] URLForResource:@"google.com" withExtension:@"cer"];
    NSData *googleCertificateData = [NSData dataWithContentsOfURL:googleCertificateURL];
    AFHTTPSessionManager *manager = [[AFHTTPSessionManager alloc] initWithBaseURL:[NSURL URLWithString:@"https://github.com/"]];
    [manager setResponseSerializer:[AFHTTPResponseSerializer serializer]];
    manager.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeCertificate withPinnedCertificates:[NSSet setWithObject:googleCertificateData]];
    [manager
     GET:@"AFNetworking/AFNetworking"
     parameters:nil
     success:^(NSURLSessionDataTask * _Nonnull task, id  _Nullable responseObject) {
         XCTFail(@"Request should fail");
         [expectation fulfill];
     }
     failure:^(NSURLSessionDataTask * _Nullable task, NSError * _Nonnull error) {
         XCTAssertEqualObjects(error.domain, NSURLErrorDomain);
         XCTAssertEqual(error.code, NSURLErrorServerCertificateUntrusted);
         [expectation fulfill];
     }];
    [self waitForExpectationsWithCommonTimeoutUsingHandler:nil];
    [manager invalidateSessionCancelingTasks:YES];
}
```


@0xced could you clean it up and get it ready for merge? :beers:

kcharwood:
Cleaned up and got test passing in #3191 

Thanks!

PR #3171, opened 2015-11-20
Reachability support IPv6.

SAMUKEI:
[Apple Document(Supporting IPv6 DNS64/NAT64 Networks)]
(https://developer.apple.com/library/prerelease/ios/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/UnderstandingandPreparingfortheIPv6Transition/UnderstandingandPreparingfortheIPv6Transition.html#//apple_ref/doc/uid/TP40010220-CH213-SW26)

PR #3174, opened 2015-11-23
Added support for IPv6 to Reachability

kcharwood:
Added IPv6 support for Reachability based on the work from @SAMUKEI in #3171

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.h
+++ b/AFNetworking/AFNetworkReachabilityManager.h
@@ -100,7 +100,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Creates and returns a network reachability manager for the socket address.
 
- @param address The socket address (`sockaddr_in`) used to evaluate network reachability.
+ @param address The socket address (`sockaddr_in6`) used to evaluate network reachability.
 
  @return An initialized network reachability manager, actively monitoring the specified socket address.
  */
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -144,10 +144,17 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
 
 + (instancetype)manager
 {
+#if (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 90000) || (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 101100)
+    struct sockaddr_in6 address;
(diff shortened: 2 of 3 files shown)


mkantor:
Not sure if this is the best place to ask, but I couldn't find a solid answer when googling and GitHub makes it easy to ask publicly:

Does this changeset imply that iOS apps using AFNetworking 2.x's reachability manager could incur the wrath of Apple on [2016-06-01, when IPv6-only support will be required](https://developer.apple.com/news/?id=05042016a)? Does AFNetworking 2.x work correctly in IPV6-only networks?

sunuslee:
@mkantor 
i want this as well:)

abournier:
Same question for me. Does AFNetworking 2.x work correctly in IPv6-only networks?

SamehDamuni:
Same question but for AFNetworking 3.x, does it work correctly in IPv6-only networks?

metaQ:
Same question for me.

abournier:
I have created a NAT64 sharing internet connection as Apple recommended to test on IPv6 only, and all seems to work for my app.

sergey-nikolaev-paycasso:
@abournier you mean the 2.x version of AFNetworking?

abournier:
Yes, the 2.6.x.

ryancrosby:
@abournier there also is some discussion related to this in #3496 related to AFNetworking 2.x supporting IPv6, probably better to keep discussion around version 2.x there rather than here which is focusing on 3.x.

abournier:
Thanks for the info. I will post on the other thread.

cezheng:
As apple has updated the [Official Reachability sample code](https://developer.apple.com/library/ios/samplecode/Reachability/Listings/Reachability_Reachability_m.html#//apple_ref/doc/uid/DTS40007324-Reachability_Reachability_m-DontLinkElementID_9), and claims that the class fully supports IPv6. But when we look at the code, it is still using `sockaddr_in` in the `reachabilityForInternetConnection` method, which makes me kind of confused.

@kcharwood However this PR actually uses ``sockaddr_in6` on iOS 9 or later. Will such implementation fail to work correctly if the device happens to be connecting to a IPv6 only network?

hei9gag:
It seems the Reachability function is not compatible to iOS 7. The function returns not reachable status even I have connected to a IPv6 network. 

Also I found that reachabilityDidChange is slightly delay to trigger in iOS 8 (compare to iOS 9) when I try to connect to a IPv6 network from no network status. Is this a bug? Many thanks.

kenmaz:
@cezheng 
https://forums.developer.apple.com/message/137546#137546
Apple's engineer say `0.0.0.0` address is treated as special token, it can be use for checking local connection check.Therefore `sockaddr_in` / `AF_INET` works good in IPv4 and Pv6 Network.

PR #3183, opened 2015-11-30
documenting signedness

Coeur:


kcharwood:
Thanks @Coeur. Would you be able to recreate this PR targeted at the 3.0.0 branch?

kcharwood:
Actually I just realized this is only for AFURLConnectionOperation, which we have now deprecated.

Thanks for the PR, but we're only taking critical patches going forward for AF 2.x

:beers:

Coeur:
Adding to documentation wouldn't have been incompatible with next critical patch release.

PR #3193, opened 2015-12-04
Split the iOS and OS X AppDelegate classes in the Example Project

0xced:
The only thing they share is their name.


kcharwood:
:beers:

PR #3202, opened 2015-12-08
add acceptableContentTypes "text/plain"

qiuyuntao:


kcharwood:
Thanks for the PR.

If this were to be merged, this would have the side effect of declaring all `text/plain` content types as capable of being parsed by the JSON serializer, which we know is that the case.

For your setup, you should use the `addAcceptableContentType` method on the response serializer to allow `text/plain` for your special case, and if you are in control of the server, you should change the type to properly represent the JSON object that is being returned.

:beers:

PR #3204, opened 2015-12-08
Remove testing from command line in readme

kyle-fox:
Looks like this functionality was removed when the Rakefile was deleted in e0815da


kcharwood:
:beers:

PR #3205, opened 2015-12-08
Task creation feature (logic improved for non-buggy versions of NSURLSession)

nikitahils:
Non-blocking task creation feature on iOS 8 + versions with support of iOS 7 version with additional logic of fixing Apple bug in NSURLSession.

More about:
Open Radar: http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
Issue: https://github.com/AFNetworking/AFNetworking/issues/2093

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -23,9 +23,9 @@
 #import <objc/runtime.h>
 
 #ifndef NSFoundationVersionNumber_iOS_8_0
-#define NSFoundationVersionNumber_With_Fiexed_5871104061079552_bug 1140.11
+#define NSFoundationVersionNumber_With_Fixed_5871104061079552_bug 1140.11
 #else
-#define NSFoundationVersionNumber_With_Fiexed_5871104061079552_bug NSFoundationVersionNumber_iOS_8_0
+#define NSFoundationVersionNumber_With_Fixed_5871104061079552_bug NSFoundationVersionNumber_iOS_8_0
 #endif
 
 static dispatch_queue_t url_session_manager_creation_queue() {
@@ -39,7 +39,7 @@ static dispatch_queue_t url_session_manager_creation_queue() {
 }
 
 static void url_session_manager_create_task_safely(dispatch_block_t block) {
-    if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fiexed_5871104061079552_bug) {
+    if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fixed_5871104061079552_bug) {
         // Fix of bug
         // Open Radar:http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
         // Issue about:https://github.com/AFNetworking/AFNetworking/issues/2093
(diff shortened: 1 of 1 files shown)


nikitahils:
Hello!

I think we should make some refactor of "NSFoundationVersionNumber_With_Fiexed_5871104061079552_bug" macros.
I choose this name, because of combination NSFoundationVersionNumber variable and bug identifier on Open Radar. May be it looks ugly, but it will be clean for people who read AFNetworking source code, i tried to explain clearly, explicitly - what's going on in this part of code.
So i repeat, that we should make some refactor of this place, but we have to clearly tell people that it's a hacky tick of fixing apple bug.

Thank you!

AFNetworking/AFURLSessionManager.m
```diff
@@ -22,6 +22,12 @@
 #import "AFURLSessionManager.h"
 #import <objc/runtime.h>
 
+#ifndef NSFoundationVersionNumber_iOS_8_0
+#define NSFoundationVersionNumber_With_Fiexed_5871104061079552_bug 1140.11
```
kcharwood: Could you changed these to `_Fixed_`?

kcharwood:
Is there any test coverage we can provide at all here?

nikitahils:
Deployment target for AFNetworking iOS Tests is 8.0, so we can't test case of iOS 7.0 (first 'if' statement). It will be simple call of block() on the same thread. 

Mac OS X 10.9 + & watchOS 2 + & tvOS 9 + (Other min. requirements) have NSFoundationVersionNumber value greater than value for NSFoundationVersionNumber_iOS_8_0, so it will be simple call of block() on the same thread.

nikitahils:
So this update give us an ability to feel free to create tasks and don't worry about blocking caller thread, at the same time we protected from NSURLSession creation task bug on iOS 7.

kcharwood:
Should we build a test that matches that open radar?

```
Steps to Reproduce:
1. Dispatch 100 data tasks on the default priority queue.
2. Monitor the taskIdentifiers returned by the task.
3. Observe that they are occasionally the same — likely, since the queue is concurrent, this is when two tasks end up getting created on different threads around the same time.
```

That would at least give me confidence in the patch moving forward, especially given I now have the test suite running on every version of iOS since 8.0.

AFNetworking/AFURLSessionManager.m
```diff
@@ -32,6 +38,17 @@ static dispatch_queue_t url_session_manager_creation_queue() {
     return af_url_session_manager_creation_queue;
 }
 
+static void url_session_manager_create_task_safely(dispatch_block_t block) {
+    if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fixed_5871104061079552_bug) {
+        // Fix of bug
+        // Open Radar:http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
+        // Issue about:https://github.com/AFNetworking/AFNetworking/issues/2093
+        dispatch_async(url_session_manager_creation_queue(), block);
```
kcharwood: This probably needs to be a dispatch_sync, since that what you are replacing in the older methods.

nikitahils:
@kcharwood, Thank you for review a lot! Yes, of course, it's definitely dispatch_sync call.

About test: It's a good idea to add test like in open radar, which target? Only for iOS? Or add to all targets (+ mac os x, tvOS)?

kcharwood:
I'd drop it in the `AFURLSessionManagerTests` class, it should get picked up by all the targets there.

kcharwood:
Will merge with #3208 


:beers: !

PR #3208, opened 2015-12-10
Improved task creation performance for iOS 8+

kcharwood:
Taking #3205 and adding a test to it.

Thanks @nikitahils!


kcharwood:
See #2093 for original issue.

Issue #2859, opened 2015-07-27
crash in AFURLSessionManager didfinishdownloading delegate method


bastianh:
Hi,
our app is crashing sometimes when we move files and doing a download at once. After some debugging I found the following line crashing with EXC_BAD_EXEC:

``` 
#pragma mark - NSURLSessionDownloadDelegate

- (void)URLSession:(NSURLSession *)session
      downloadTask:(NSURLSessionDownloadTask *)downloadTask
didFinishDownloadingToURL:(NSURL *)location
{
...
[[NSFileManager defaultManager] moveItemAtURL:location toURL:fileURL error:&error];
...
}
```
the app crashes if an other thread (or the mainthread) moves a file at the same time. 

I suggest using a own instance of NSFileManager in the library, as mentioned in the class description:

> In iOS and Mac OS X v 10.5 and later you should consider using [[NSFileManager alloc] init] rather than the singleton method defaultManager. Instances ofNSFileManager are considered thread-safe when created using [[NSFileManager alloc] init].

kcharwood:
Thanks @bastianh. Would you be interesting in putting together the PR?

osrufung:
Same problem here under iOS8 SDK + iPhone Simulator. Any idea?

kcharwood:
I looked into this recently.

The documentation states:

> The methods of the shared NSFileManager object can be called from multiple threads safely. However, if you use a delegate to receive notifications about the status of move, copy, remove, and link operations, you should create a unique instance of the file manager object, assign your delegate to that object, and use that file manager to initiate your operations.

Do you have any data showing that isn't the case?

PR #3211, opened 2015-12-10
Using thread safe NSFileManager objects instead of shared instance

kyle-fox:
Referenced here: https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSFileManager_Class/
Issue: #2859

### DIFF
--- a/AFNetworking/AFURLRequestSerialization.m
+++ b/AFNetworking/AFURLRequestSerialization.m
@@ -707,7 +707,8 @@ NSTimeInterval const kAFUploadStream3GSuggestedDelay = 0.2;
         return NO;
     }
 
-    NSDictionary *fileAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:[fileURL path] error:error];
+    NSFileManager *fileManager = [[NSFileManager alloc] init];
+    NSDictionary *fileAttributes = [fileManager attributesOfItemAtPath:[fileURL path] error:error];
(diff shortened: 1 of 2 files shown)


kcharwood:
I looked into this recently.

The documentation states:

> The methods of the shared NSFileManager object can be called from multiple threads safely. However, if you use a delegate to receive notifications about the status of move, copy, remove, and link operations, you should create a unique instance of the file manager object, assign your delegate to that object, and use that file manager to initiate your operations.

Do you have any data showing that isn't the case?

kyle-fox:
You're right - it looks like the singleton was not thread-safe in the past and seems to be now, with some caveats. Using the unique instances might be the safer choice at the expense of allocating multiple objects since it's always been the thread-safe option, but I'll leave that choice to you. I'd be curious to see if this fix solved the crashes in #2859 - I haven't been able to replicate it.

kyle-fox:
I'll have to test this to be sure, but the user could be implementing `NSFileManagerDelegate` to move files on multiple threads using the `defaultManager`, which Apple doesn't recommend.

>This method always returns the same file manager object. If you plan to use a delegate with the file manager to receive notifications about the completion of file-based operations, you should create a new instance of NSFileManager (using the init method) rather than using the shared object.

While this crash wouldn't be caused directly by anything in AFNetworking, the crash might be prevented from occurring within AFN by allocating its own instance.

PR #3214, opened 2015-12-11
Fixed race condition with image downloader when starting/cancelling/starting the same request

kcharwood:
See https://github.com/Alamofire/AlamofireImage/issues/55 for details

@cnoon can you give this a review?


cnoon:
Looks good dude. 👍🏼

kcharwood:
I realized this approach didn't yet solve the same problem for the UIButton category, which is more complex because we have 8 different possible active downloads. I put up #3215 as an alternative solution. Thoughts on the API addition there?

kcharwood:
Closing this. We're going to go with #3215

PR #3215, opened 2015-12-11
Fixed race condition with ImageView/Button image downloading when starting/cancelling/starting the same request

kcharwood:
Alternate Solution for #3214 

See https://github.com/Alamofire/AlamofireImage/issues/55 for details

@cnoon which one do you think is better? I think I like approach better, but it does add a new API method to the downloader. Should still be backwards compatible.


cnoon:
Yeah buddy...definitely a more scalable approach. It also solves the problem of having to pass the `RequestReceipt` to the `completion` closure to validate the `receiptID`. 👍🏼

PR #3220, opened 2015-12-14
Button image changed instead of background image

eofs:
Background image did not change when `setBackgroundImageForState` called as `setImage` was used instead of `setBackgroundImage`.


kcharwood:
Thanks so much! :beers:

PR #3222, opened 2015-12-14
Fixed crash in `AFURLSessionManager` when resuming download task.

lembacon:
As long as there is already a `downloadTaskDidResume` block, this invocation should be removed.


kcharwood:
Thanks!

PR #3234, opened 2015-12-17
Potential leak of an object stored into 'reachability'

yulingtianxia:
Call to function 'SCNetworkReachabilityCreateWithName' returns a Core Foundation object with a +1 retain count
Object leaked: object allocated and stored into 'reachability' is not referenced later in this execution path and has a retain count of +1


kcharwood:
I believe this is handled by this [line](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFNetworkReachabilityManager.m#L167)

PR #3246, opened 2015-12-24
add retryTimes and retryFilter

heroims:
It's easy for us to retry..

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.h
+++ b/AFNetworking/AFHTTPSessionManager.h
@@ -75,6 +75,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface AFHTTPSessionManager : AFURLSessionManager <NSSecureCoding, NSCopying>
 
+@property (nonatomic, assign) NSInteger retryTimes;
+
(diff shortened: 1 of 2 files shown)


kcharwood:
I'm not sure how I feel about adding a retry mechanism to the core library. Typically, this is behavior I've seen offloaded in a subclass of your session manager. It also hasn't been an often requested feature.

Thoughts @0xced @cnoon @kylef 

Would love to hear from the rest of the community as well.

cnoon:
I feel pretty strongly about keeping this type of behavior out of the core library. It seriously complicates things. This PR only considers data tasks, not upload or download tasks. Additionally, retry policies can essentially DDoS your servers when you have thousands or millions of users. They need to come with a progressive back off policy in order to avoid it. On-top of that, most back off policies need to be customized for each endpoint. Most of them all need to be treated differently depending on your use case.

Hence, I think it should be left out of AFN and AF core.

kcharwood:
You did a much better job of summing up my feelings :)

Thanks for the PR, but I'm going to close this one out.

kylef:
I agree with what @cnoon said :+1:.

PR #3253, opened 2015-12-29
Removed workaround for NSJSONSerialization bug that was fixed in iOS 7

0xced:
The NSJSONSerialization bug with Unicode character escape codes has been fixed in iOS 7, see http://openradar.appspot.com/radar?id=2397401#ag9zfm9wZW5yYWRhci1ocmRyFAsSB0NvbW1lbnQYgICAgIDdwwgM

Removing the NSData -> NSString -> NSData roundtrip is also likely a nice performance boost.


0xced:
For reference, this workaround was introduced in #660.

kcharwood:
:beers:

PR #3254, opened 2015-12-30
Add error parameter to the examples of AFHTTPRequestSerializer on README.md

gunyarakun:
Now AFHTTPRequestSerializer always requires error parameter, but the examples on README.md don't follow the change.


kcharwood:
Thanks!

PR #3259, opened 2016-01-04
Changing confusing URL in Readme for data task

corymsmith:
This URL threw me off when I first read it, figured this way will reduce confusion for others.


kcharwood:
I'd rather the URL be something that returns JSON, since managers are configured by default to handle that.

Could you change it to `http://httpbin.org/get`?

Thanks!

corymsmith:
Good point. Will do!

kcharwood:
:beers:

PR #3262, opened 2016-01-04
Modify function `AFJSONObjectByRemovingKeysWithNullValues`

huuang:
The original implement of function AFJSONObjectByRemovingKeysWithNullValues can not remove any keys with null values. The new one will remove keys which have null value, or empty string value, or empty array value, or empty dictionary value.


kcharwood:
It looks like the test suite is not passing with this change. Could you please get the tests passing, and add any new tests for your test case? Thanks!

PR #3187, opened 2015-12-01
New Progress Reporting API using `NSProgress`

kcharwood:
These changes expose a simpler, block based progress reporting API, while maintaining the ability for more advanced features using KVO on NSProgress.

Issues #3158 and #3163 got me thinking that progress **_was easier_** to use with old NSURLConnection based API's, and we shouldn't force anyone to use KVO for progress updating, so I decided to take a stab at implementing a block based approach, while also streamlining the accessors for individual upload and download progress.

Note that `NSURLSessionTask` sees progress as two distinct items: upload progress and download progress, as reported by `countOfBytesExpectedToReceive`, `countOfBytesReceived`, `countOfBytesExpectedToSend`, and `countOfBytesSent`. Therefore, I've decided to track these two things independently with two distinct progress objects, managed by the internal `AFURLSessionManagerTaskDelegate`.

### API Changes

* Additional convenience API's in `AFHTTPSessionManager` exposing download progress block for GET, and upload progress block for POST.
* Streamlined API in `AFURLSessionManager` allowing upload and/or download NSProgress objects to be retrieved for any task.
* Additional convenience API's in `AFURLSessionManager` exposing download and upload progress blocks for data task, download progress block for download tasks, and upload progress block for upload tasks.

### Open Questions
* Should the methods not containing the progress blocks in `AFHTTPSessionManager` be deprecated and/or removed? For example, there are now four POST methods in that API, rather than two as before.
  * I am leaning towards removing them completely and listing it as an upgrade note. 
* Currently, the progress blocks are called on the session queue, not the main queue. I tinkered with dispatching them to the main queue, but it resulted in "duplicated" progress reporting because the progress object was updated again on the session queue before the main queue was called, making it seem like there actually wasn't a change in progress. Should we care about this?
   * Progress Updated to 2%. Dispatched to main queue
   * Progress Updated to 5%. Dispatched to main queue
   * Dispatch called on main queue. Reports 5% progress since its a _reference_ to the current progress
   * Dispatch called on main queue. Reports 5% progress since its a _reference_ to the current progress 

@0xced @cnoon @jshier @kylef would love your eyes on this. Probably the biggest code change so far in the 3.0.0 branch. I think I got all the KVO management setup internally correctly, but would appreicate a review.

PR #3272, opened 2016-01-09
Fixes #3271, Progress blocks not called on main thread

### LINKED ISSUE #3271: Progress blocks not called on main thread
Similar to issue #411 for the operation based APIs, the progress blocks for the session APIs aren't being called on the main thread.

For session based APIs, the KVO observation callbacks are happening on Apple's work queue (eg `com.apple.NSURLSession-work`), so that is the queue the upload and download progress blocks are being called on.

Instead, it would be good to have these calls happen on the same work queue and group as the completion callbacks. This would be in line with the operation based APIs.

I discovered this problem when refactoring a codebase to use the new API, and hit a threading problem when calling into UIKit from a background queue. I suspect others will encounter this as they move to the session based API.

wooster:


kcharwood:
Thanks for the PR!

This is a tricky part to the NSProgress API.

Check out my reasoning in #3187 for why this is the case. It's also called out in the documentation for all of those methods. Basically, if you dispatch out to the main queue, and pass the NSProgress reference, the reference may be updated again before it runs on the main queue, leading to confusion and the appearance of nothing being updated.

By returning it on the session queue, I leave flexibility in to allow the caller to get every update, and then determine what to do with it (in the case it needs to be vended to the main queue for UI updates).

I'm open to hearing feedback from the community, but that's where I've landed so far.

wooster:
@kcharwood: I'm not a big fan of leaving land mines like this in API.

Two alternatives:

* Un-deprecate the methods without the progress blocks. As it stands, we've now got potential threading problems all over our code base, waiting to be misused.
* Construct new NSProgress objects for each update, and send those across the completion queue.

kcharwood:
It's something I've _really_ struggled with. I don't think creating new NSProgress objects is the best option for how NSProgress is designed (give the use of child progresses and things like that).

I would say the options are:

1) Keep it as is, and instruct through documentation and proper naming.
2) Convert the block to use primitive objects, and simply pass them back and let the caller do the math (`bytesReceived`, `expectedBytesReceived`, etc)

The first option gives us the ability to leverage the power of NSProgress (still an underused object in the iOS SDK), while forcing us to manage main queue dispatching ourselves. The second forces us to lose the power of NSProgress, but get a better expected callback.

Note that NSURLSession does call you back on the session queue for those updates, so its a pattern that does follow how NSURLSession works.

Unfortunately I'm left with two options that I don't 100% like :(

wooster:
My vote would be to un-deprecate the API that doesn't include the progress block parameters.

I'll keep looking for another solution, though, because having this kind of land mine in large projects isn't ideal.

kcharwood:
I'll do some more thinking on this front. I view it as a "less" of a land mine because its the same behavior NSURLSession would give you if you were use it directly. I agree its not optimal though.

PR #3275, opened 2016-01-12
add @"text/html" will solve some special request

XYXiaoYuan:
some json will include the type @"text/html",if here not add this,this type request can't return any data for us.
I have encountered this error,NSLocalizedDescription=Request failed: unacceptable content-type: text/html.
so I hope will check it out ,and accept it soon!

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -223,7 +223,7 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
         return nil;
     }
 
-    self.acceptableContentTypes = [NSSet setWithObjects:@"application/json", @"text/json", @"text/javascript", nil];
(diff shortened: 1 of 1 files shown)


XYXiaoYuan:
add @"text/html" will solve some special request
some json will include the type @"text/html",if here not add this,this type request can't return any data for us.
I have encountered this error,NSLocalizedDescription=Request failed: unacceptable content-type: text/html.
so I hope will check it out ,and accept it soon!

kylef:
In future, please check for similar pull requests (to prevent duplicate work and review). This is very similar to #2447, #2639, #2871.

> This is not the correct solution. If a server is not sending a non-standard or incorrect content type, set the `acceptableContentTypes` property as necessary.

> Adding non-standard acceptable content types is a slippery slope. Not all HTML or javascript is XML compatible, so dropping this in here potentially causes other issues. For your use, you can easily update the acceptable content types for each serializer, so there is no need to patch it here.

Or better yet, fix your server/API to use the correct content-types for JSON.

XYXiaoYuan:
Thank u,I'm sorry I didn't check the similar requests.This is my frist pull requests,I'm look forward to the AFNetworking will be better and better.

PR #3284, opened 2016-01-13
Update AFHTTPSessionManager.h

revolter:
Mark progress block as nullable.

There may be other blocks that require this, but I'm not sure which ones. I am still migrating to 3.0 and I didn't use AFNetworking much to know it better.

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.h
+++ b/AFNetworking/AFHTTPSessionManager.h
@@ -158,7 +158,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable NSURLSessionDataTask *)GET:(NSString *)URLString
                             parameters:(nullable id)parameters
-                              progress:(nullable void (^)(NSProgress *downloadProgress)) downloadProgress
+                              progress:(nullable void (^ __nullable)(NSProgress *downloadProgress)) downloadProgress
                                success:(nullable void (^)(NSURLSessionDataTask *task, id _Nullable responseObject))success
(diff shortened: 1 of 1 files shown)


kcharwood:
The progress block is already nullable. Looking at the swift generated header, you can see the progress parameter is an optional, meaning it is nullable.

```Swift
    /**
     Creates and runs an `NSURLSessionDataTask` with a `GET` request.
    
     @param URLString The URL string used to create the request URL.
     @param parameters The parameters to be encoded according to the client request serializer.
     @param progress A block object to be executed when the download progress is updated. Note this block is called on the session queue, not the main queue.
     @param success A block object to be executed when the task finishes successfully. This block has no return value and takes two arguments: the data task, and the response object created by the client response serializer.
     @param failure A block object to be executed when the task finishes unsuccessfully, or that finishes successfully, but encountered an error while parsing the response data. This block has no return value and takes a two arguments: the data task and the error describing the network or parsing error that occurred.
    
     @see -dataTaskWithRequest:uploadProgress:downloadProgress:completionHandler:
     */
    public func GET(URLString: String, parameters: AnyObject?, progress downloadProgress: ((NSProgress) -> Void)?, success: ((NSURLSessionDataTask, AnyObject?) -> Void)?, failure: ((NSURLSessionDataTask?, NSError) -> Void)?) -> NSURLSessionDataTask?
```

Let me know if I have misunderstood you!

:beers:

revolter:
You're right, I misunderstood something.

PR #3286, opened 2016-01-14
I found a serious bug about UIWebView+AFNetworking

snowimba:
demo about this bug


kcharwood:
Thanks so much! I've patched this in #3289 

:beers:

PR #3290, opened 2016-01-14
Bumping copyright to 2016

kcharwood:


kylef:
What about the [LICENSE file](https://github.com/AFNetworking/AFNetworking/blob/master/LICENSE#L1) too?

kcharwood:
Good catch :beers:

Issue #2797, opened 2015-06-19
Server response with status code 205 leads to timeout


bennibau:
Our Backend sends at a certain call the response 205 without content (according to HTTP status code specification)
The app waits until the timeout has run out and returns with a timeout. Is the call waiting for content which will not be sent?

kcharwood:
[By default](https://github.com/AFNetworking/AFNetworking/blob/2.5.4/AFNetworking/AFURLResponseSerialization.m#L98), code 205 is an acceptable response code and behaves just like any other 2xx code in AFNetworking

dgustafsson-dsi:
I am seeing the same behaviour. 
A response with code 205 does not return the success block - The request times out and we get a timeout in the failure block. 
I see that 205 is an acceptable response according to the above. But if it was not acceptable the failure block would be invoked immediately, it would not time out.

There's also a SO post regarding the same here
http://stackoverflow.com/questions/28031209/stuck-in-a-request-with-response-205

kcharwood:
What is being returned for your response content type?

dgustafsson-dsi:
The AFHTTPRequestOperation's response object is 'nil'.

bennibau:
Yep as I said in June, the framework does not come back to the call. It times out even if the server responds correctly.

kcharwood:
Sorry. I meant your actually http response headers. Not the cocoa object.

bennibau:
sorry I cannot provide details on that anymore since it was long time ago. We changed our backend using another responsecode for that call.

dgustafsson-dsi:
The response headers, as in? 
operation.response.allHeaderFields
operation.response is 'nil'
so there are no headers. There is no response.

kcharwood:
Sorry. I'm saying if you were to hit with a web request app, or Charles, or postman, etc, what is the full response and all the headers being returned by the server. You won't see that in AF because the response object is nil.

You could dig deeper into NSURLSession and get that information as well.

Alternatively, do you have a public URL i could hit?

dgustafsson-dsi:
Ah, got you.
Using Charles I get these response headers

HTTP/1.1 205 Reset Content
Server: Apache-Coyote/1.1
Date: Thu, 14 Jan 2016 16:04:58 GMT

kcharwood:
Gotcha. So I'm assuming the problem here is there is no response content type, which causes the response serializer to fail.

I'll need to take a look at this.

kcharwood:
[...]

PR #3292, opened 2016-01-14
Fixed status code 204/205 handling

kcharwood:
If a server returned a 204/205 with no response content type and no response data, the response serializer would fail. This patch allows response serialization to continue without error if no response content type header is defined, and there are zero bytes of response of data.

If the server returns a response content type header OR response data, the serializer will respect the old serialization logic. This means that if a response content type is listed in the header, we are expecting some type of data response.

This patches #2797

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -115,7 +115,9 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
     NSError *validationError = nil;
 
     if (response && [response isKindOfClass:[NSHTTPURLResponse class]]) {
-        if (self.acceptableContentTypes && ![self.acceptableContentTypes containsObject:[response MIMEType]]) {
+        if (self.acceptableContentTypes && ![self.acceptableContentTypes containsObject:[response MIMEType]] &&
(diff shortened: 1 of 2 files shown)


kcharwood:
cc @kylef @cnoon @jshier @0xced 

Any thoughts here on this patch?

cnoon:
Hey @kcharwood, sorry for not getting back to you sooner. 

It's a bit difficult for me to tell exactly what all the moving pieces are here in this logic. I made a very similar type change to Alamofire about 4 months ago to support 204 and 205 cases properly. I stopped validating the content type if there was no data returned which you can see [here](https://github.com/Alamofire/Alamofire/blob/master/Source/Validation.swift#L131-L165). The reason was that iOS will "infer" a content-type even if one isn't returned by the server. I was seeing "text/plain" being reported as the `MIMEType` even though nothing was returned. Because of this behavior, the only consistent way around the problem that I could find was to just ignore the content type validation if no data was returned. That made things simple and seemed like a reasonable tradeoff.

I can't tell if that's exactly what you're doing here, but that ended up being my reasoning.

kcharwood:
My check here is adding an additional check to prevent response type validation if there is no MIME type and the response data length == 0.

The inferred content type however may throw a wrench in that?

Issue #3282, opened 2016-01-13
iOS8.0, 9.0 is OK, but progress block not called at all，iOS7.0


emilyZhouwm:
iOS8.0, 9.0 is OK, but progress block not called at all，iOS7.0

    AFURLSessionManager *manager = [EHDownloadNetwork sharedInstance].manager;
    downloadTask.task = [manager downloadTaskWithRequest:request
              progress:^(NSProgress * _Nonnull downloadProgress) {
                  if (progressBlock) {
                      dispatch_async(dispatch_get_main_queue(), ^{
                          progressBlock(downloadProgress.fractionCompleted);
                      });
                  }

kcharwood:
Can you provide more information about the response the server is sending? AFNetworking hooks into the NSURLSession progress update properties, which rely on a response to properly format the expected amount of bytes to be returned by the response.

kcharwood:
It looks like you have chimed in on #3270, so we'll direct comments there.

:beers:

PR #3294, opened 2016-01-15
Fixed issue where download progress would not be reported in iOS 7

emilyZhouwm:
iOS8.0, 9.0 is OK, but progress block not called at all，iOS7.0 #3282


kcharwood:
Thanks for tracking this down @emilyZhouwm.

I didn't think about iOS 7 impact, but given the class cluster implementation that [differs](https://github.com/AFNetworking/AFNetworking/pull/2702#issuecomment-100413218) in iOS 7, its not surprising this is needed. I wish we had an iOS 7 test target to verify this fix, but unfortunately we do not :(.


Thanks!

emilyZhouwm:
^_^We adapted to the minimum to iOS7, so will test the equipment retained iOS7 and iOS8 ipod touch operating system.Now is only 10% of users inside iOS7

PR #3299, opened 2016-01-17
Tighten the description of the progress parameter

fulldecent:


fulldecent:
Related code which guarantees the tighter description is true

https://github.com/AFNetworking/AFNetworking/blob/aa9b5f59aa4927c1ed0415ef7b5115662d17b226/AFNetworking/AFURLSessionManager.m

kcharwood:
I'm confused by the changes here. The progress block is called multiple times through the lifetime of the lifetime of the response. What scenario are you trying to document?

fulldecent:
I use the progress block and not the completion block.

My expectation is that there will be exactly one callback where progress = 1.0, indicating that it is done.

kcharwood:
The progress block should be called multiple times if your response properly declares a `Content-Length`.

If your response does not include that information, then I believe it will only be called when it is finished.

fulldecent:
Right. Either way:

If has content-length

-> Progress with 1.0 is called once after complete

It does not have content-length

-> Progress with 1.0 is called once after complete

kcharwood:
Right. But you're documentation change to me is confusing:

> This block is guaranteed to be called exactly once if the task finished successfully.

The block is called multiple times for progress reporting. This note seems to make it seem that it will only be called once.

kcharwood:
Also note that you only get one 1.0 progress callback because of my notes in #3187 and having the progress callback happen on the main queue.

If that ever changes, and the progress update is dispatched to the main queue, it's possible you may see multiple 1.0 callbacks.

I would recommend relying on the success block, which is _guaranteed_ to only be called once, and completely within our control.

fulldecent:
@kcharwood Sorry about this and thanks for pointing out. I have corrected the text.

aozhimin:
@fulldecent , does this pull request solve the problem which I have encountered, Thank You!

fulldecent:
@aozhimin Yes, this resolves our discussion at https://github.com/AFNetworking/AFNetworking/pull/3187#issuecomment-172413036

aozhimin:
@fulldecent , thank you for telling me

kcharwood:
Technically, all unique progress values should be reported once. Something about the way this is worded still seems off to me, and like I said earlier, I would rely on the success block, not the progress block, for tracking completion.

kcharwood:
I'm going to close this for now. I will revisit it if we see any more confusion/reported issues around this in the future.

:beers:

Issue #3301, opened 2016-01-19
Download progress on UIImageView+AFNetworking category


knutigro:
It would be nice to have access to the download progress when using the UIImageView+AFNetworking category. 

I did add a question here:  http://stackoverflow.com/questions/34744024/afnetworking-3-0-setimagewithurlrequest-download-progress 

and I also did some experiments by copying some code from `UIProgressView+AFNetworking.h`. The result is here: https://gist.github.com/knutigro/85d3463f06f660d7a67d . My category crashes  in some situations though, so it still need more work :-( 

An other ideas? Are you open for pull requests on this issue?

kcharwood:
It should be possible to expose the progress update. I would certainly be open to a pull request. It should be accessible from the manager on the image downloader, and just needs to be exposed back up through the image view interface.

knutigro:
@kcharwood  Here is my pullrequest. 

I did notice that in both the success block and in the failure block you check against the `activeImageDownloadReceipt.receiptID` before actually calling the block. I did not understand why that should be needed in the progress block so I left it out. Please correct me if I am wrong.

The method I talk about is `- (void)setImageWithURLRequest:(NSURLRequest *)urlRequest
              placeholderImage:(UIImage *)placeholderImage  downloadProgress:(nullable void (^)(NSProgress *downloadProgress)) downloadProgressBlock success:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, UIImage *image))success failure:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, NSError *error))failure`

and the line of code I reference to is:

`if ([strongSelf.af_activeImageDownloadReceipt.receiptID isEqual:downloadID])`...

PR #3306, opened 2016-01-21
Add image download progress on UIImageView+AFNetworking category

knutigro:
Fixes https://github.com/AFNetworking/AFNetworking/issues/3301


knutigro:
I did notice that in both the success block and in the failure block you check against the activeImageDownloadReceipt.receiptID before actually calling the block. I did not understand why that should be needed in the progress block so I left it out. Please correct me if I am wrong.

The method I talk about is - (void)setImageWithURLRequest:(NSURLRequest *)urlRequest placeholderImage:(UIImage *)placeholderImage downloadProgress:(nullable void (^)(NSProgress *downloadProgress)) downloadProgressBlock success:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, UIImage *image))success failure:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, NSError *error))failure

and the code I reference to is:

                      __strong __typeof(weakSelf)strongSelf = weakSelf;
                       if ([strongSelf.af_activeImageDownloadReceipt.receiptID isEqual:downloadID]) {
                           if (success) {
                               success(request, response, responseObject);
                           } else if(responseObject) {
                               strongSelf.image = responseObject;
                           }
                           [strongSelf clearActiveDownloadInformation];
                       }

kcharwood:
Looks like #3318 also has a similar change.

I have not had a chance to review either. Could you please sync this up with master to get the fix for the units tests, and could you please add test cases as well?

knutigro:
@kcharwood  I did add one test for the AFImageDownloader and one for the UIImageView category. Both are based on how you do check download progress in other places in AFNetworking. Hope it's ok like that :-)

kcharwood:
You need to make sure you support multiple callers requesting the same image URL, and _both_ getting the progress callback. I don't believe your implementation currently handles that.

knutigro:
@kcharwood Hi again. I'ts been long time with this pullrequest. I am using my own fork in the app Foap which is a photo app with lots of activity. Therefore I also see that my fix is **not good**. I get crashes because of my changes. I have hit the wall with finding a solution. 
Here is the code which is crashing in `AFImageDownloader.m` line 254 of my fork:
```
downloadProgress:^(NSProgress * _Nonnull progress) {
   dispatch_async(self.responseQueue, ^{
       AFImageDownloaderMergedTask *mergedTask = self.mergedTasks[URLIdentifier];
       if ([mergedTask.identifier isEqual:mergedTaskIdentifier]) {
           for (AFImageDownloaderResponseHandler *handler in [mergedTask.responseHandlers copy]) {
               if (handler.progressBlock) {
                   dispatch_async(dispatch_get_main_queue(), ^{
                       handler.progressBlock(progress);
                   });
               }
           }
       }
   });
}
```
The problem was originally that I was iterating this mutable array `mergedTask.responseHandlers` and got crashes because it was modified during iteration. My solution was to create a `copy` before the iteration, but this seeme to trigger new types of crashes. One of them say:

> *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[NSArray initWithArray:range:copyItems:]: range {0, 1} extends beyond bounds for empty array' 

I wonder if it would be possible to get some hints? Maybe engage some other more experienced developers? Would you have time or maybe you could ping someone else into this thread?

Ping @ncreated ;-)

jshier:

PR #3315, opened 2016-01-26
Fixed static analyzer warnings on AFNetworkReachabilityManager

SlaunchaMan:
The current version of `AFNetworkReachabilityManager` fixes some static analyzer warnings by using `#ifndef __clang_analyzer__` to hide the errors. This PR removes that conditional code in favor of fixing the analyzer issue.

Instead of bridging the reachability reference to ARC, this PR sets the type of `networkReachability` to `SCNetworkReachabilityRef`, which not only allows for cleaner memory management as far as the analyzer is concerned, but also gets rid of a lot of unnecessary bridging further into the file.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -176,11 +176,11 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
 }
 
 - (void)dealloc {
+    [self stopMonitoring];
+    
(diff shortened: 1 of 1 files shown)


kcharwood:
Could you sync this back up with master to get the fixes for the expired certificate? That should allow the test suite to pass.

SlaunchaMan:
Rebased!

kcharwood:
Thanks!

Issue #3324, opened 2016-01-29
AFImageDownloader: Downloading multiple images fails when the tasks are cancelled and started again


florianschulz:
When you start downloading multiple images simultaneously, then immediately cancel these downloads and then start them again, some downloads fail with cancelled status. This seems to be tied to the value of the "maximumActiveDownloads" property of AFImageDownloader, since the number of attempted downloads has to be greater than this value in order to make some of the downloads fail. Increasing this value also solves the issue.
The following is a failing unit test that can be added to AFImageDownloaderTests to demonstrate this issue:

```objc
- (void)testThatItCanDownloadAndCancelAndDownloadAgain {
    XCTestExpectation *expectation = [self expectationWithDescription:@"no download should fail"];
    __block NSUInteger successCounter = 0;
    
    NSArray *imageURLStrings = @[
                                 @"https://static.pexels.com/photos/1562/italian-landscape-mountains-nature.jpg",
                                 @"https://static.pexels.com/photos/397/italian-landscape-mountains-nature.jpg",
                                 @"https://static.pexels.com/photos/2698/dawn-landscape-mountains-nature.jpg",
                                 @"https://static.pexels.com/photos/2946/dawn-nature-sunset-trees.jpg",
                                 @"https://static.pexels.com/photos/5021/nature-sunset-person-woman.jpg"
                                 ];
    
    for (NSString *imageURLString in imageURLStrings) {
        AFImageDownloadReceipt *receipt = [self.downloader downloadImageForURLRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:imageURLString]] success:nil failure:nil];
        [self.downloader cancelTaskForImageDownloadReceipt:receipt];
    }
    
    for (NSString *imageURLString in imageURLStrings) {
        [self.downloader downloadImageForURLRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:imageURLString]] success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull responseObject) {
            successCounter++;
            if (successCounter == [imageURLStrings count] - 1) {
                [expectation fulfill];
            }
        } failure:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, NSError * _Nonnull error) {
            [expectation fulfill];
        }];
    }
    
    [self waitForExpectationsWithCommonTimeoutUsingHandler:nil];
    
[...]

PR #3325, opened 2016-01-29
Fixed image downloading cancellation race condition

kcharwood:
Addresses #3324

### DIFF
--- a/Tests/Tests/AFImageDownloaderTests.m
+++ b/Tests/Tests/AFImageDownloaderTests.m
@@ -343,6 +343,39 @@
     XCTAssertNotNil(responseImage);
 }
 
+- (void)testThatItCanDownloadAndCancelAndDownloadAgain {
+    XCTestExpectation *expectation = [self expectationWithDescription:@"no download should fail"];
+    __block NSUInteger successCounter = 0;
+
+    NSArray *imageURLStrings = @[
+                                 @"https://static.pexels.com/photos/1562/italian-landscape-mountains-nature.jpg",
+                                 @"https://static.pexels.com/photos/397/italian-landscape-mountains-nature.jpg",
+                                 @"https://static.pexels.com/photos/2698/dawn-landscape-mountains-nature.jpg",
+                                 @"https://static.pexels.com/photos/2946/dawn-nature-sunset-trees.jpg",
+                                 @"https://static.pexels.com/photos/5021/nature-sunset-person-woman.jpg"
+                                 ];
+
+    for (NSString *imageURLString in imageURLStrings) {
+        AFImageDownloadReceipt *receipt = [self.downloader downloadImageForURLRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:imageURLString]] success:nil failure:nil];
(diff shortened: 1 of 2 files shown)


florianschulz:
Thanks for addressing this issue that fast! 👍🏻

Unfortunately the unit test example I posted within the issue #3324 is wrong. The successCount should equal the number of requests. I also removed the image URLs from the test since they were fairly large downloads. The unit test now fails again, because now neither the failure block nor the success block get called anymore of the requests whose indices exceed the value in "maximumActiveDownloads". 

The fixed unit test:

```objc
- (void)testThatItCanDownloadAndCancelAndDownloadAgain {
    AFImageDownloader *downloader = [[AFImageDownloader alloc] initWithSessionManager:self.downloader.sessionManager
                                                               downloadPrioritization:self.downloader.downloadPrioritizaton
                                                               maximumActiveDownloads:1
                                                                           imageCache:self.downloader.imageCache];
    
    XCTestExpectation *expectation = [self expectationWithDescription:@"no download should fail"];
    __block NSUInteger successCount = 0;
    
    NSArray *imageURLRequests = @[self.pngRequest, self.jpegRequest];
    
    for (NSURLRequest *imageURLRequest in imageURLRequests) {
        AFImageDownloadReceipt *receipt = [downloader downloadImageForURLRequest:imageURLRequest success:nil failure:nil];
        [downloader cancelTaskForImageDownloadReceipt:receipt];
    }
    
    for (NSURLRequest *imageURLRequest in imageURLRequests) {
        [downloader downloadImageForURLRequest:imageURLRequest success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull responseObject) {
            successCount++;
            if (successCount == [imageURLRequests count]) {
                [expectation fulfill];
            }
        } failure:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, NSError * _Nonnull error) {
            [expectation fulfill];
        }];
    }
    
    [self waitForExpectationsWithCommonTimeoutUsingHandler:nil];
    
    XCTAssertTrue(successCount == [imageURLRequests count], @"all downloads should succeed");
}
```

kcharwood:
You are right. I'll update that and then dig further.

The root of the issue here is cancelling a request and kicking it off again before the original one has a to chance to cancel, and only if that request to be cancelled is in the suspended state.

kcharwood:
Thanks @florianschulz 

I've updated the unit test to be a bit better, and improved my patch. Let me know your thoughts.

florianschulz:
Great job! It's working now for me. ✌️
Thanks again for addressing the issue so quickly. 

Just as a suggestion, I would still replace the image URLs in the test before merging it into master, since due to their download size, there is a chance of getting an expectation timeout under bad network conditions. It is also possible that the images will be removed from the server in the future. I couldn't find a good source for reliable image downloads, so I modified the test (as suggested above) to download only two images from httpbin and setting the maximumActiveDownloads to 1.

kcharwood:
Thanks for the tip. Just switched over to gravatar.

kcharwood:
Tests failed due to a certificate expiring. Need to get that one fixed first.

PR #3344, opened 2016-02-17
Fixed image memory calculation

SummertimSadness:
Fixed image memory calculation

### DIFF
--- a/UIKit+AFNetworking/AFAutoPurgingImageCache.m
+++ b/UIKit+AFNetworking/AFAutoPurgingImageCache.m
@@ -44,8 +44,8 @@
 
         CGSize imageSize = CGSizeMake(image.size.width * image.scale, image.size.height * image.scale);
         CGFloat bytesPerPixel = 4.0;
-        CGFloat bytesPerRow = imageSize.width * bytesPerPixel;
-        self.totalBytes = (UInt64)bytesPerPixel * (UInt64)bytesPerRow;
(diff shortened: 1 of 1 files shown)


SummertimSadness:
Fixed image memory calculation

cnoon:
This logic is exactly the same @SummertimSadness. Your calculation just multiplies `width` x `height` and then by `bytesPerPixel` instead of the row size which is `width` x `bytesPerPixel` then finally by the `height`. Both operations result in the same value. 

I'm going to close this PR out for now. If I'm missing something very obvious, please point out my mistake and I'll gladly re-open.

Cheers. 🍻

tewha:
Ack. Take another look.

        CGFloat bytesPerRow = imageSize.width * bytesPerPixel;
        self.totalBytes = (UInt64)bytesPerPixel * (UInt64)bytesPerRow;

Where's `imageSize.height`?

cnoon:
Doh, sometimes you look at something way too many times. Thanks @tewha!

kcharwood:
Thanks for merging this in @cnoon!

Do you recall if there were issues with tests here? I think I recall the original PR failing tests.

cnoon:
Hmmmm...let me take a look. Travis has been so messed up over the past two weeks on AF that I've stopped paying attention.

SummertimSadness:
我感觉totalBytes的计算方式不是最好的，而且dispatch_barrier_async的性能很低，我会继续优化这部分代码，非常高兴能为AF贡献

PR #3355, opened 2016-02-23
Fix - Image Cache Tests

cnoon:
This PR fixes the auto-purging image cache tests that were broken by merging #3344. I've verified the changes are accurate. The logo image is 850 x 300 px with a pixel size of 4 bytes which totals 1020000 bytes.


kcharwood:
Thanks @cnoon

Issue #3343, opened 2016-02-16
AFImageDownloader  identifier  is nil when app crashes


soulCorw:
![0bbc3f7e-ab79-4b9d-9dba-46dec3cd6a47](https://cloud.githubusercontent.com/assets/11462712/13069308/7f43db00-d4be-11e5-8801-3337ed4e118f.png)

 reason: '*** setObjectForKey: key cannot be nil'

xianwenx:
Check the URL being passed into the request, do you still have it?

soulCorw:
Yes，I've been tracking it，at the beginning of the request, URL was a nil。I suggest that if there is an assertion to URL，we'll be closer to perfection。

kcharwood:
Please see #3385.

We'll direct conversation there for now.

:beers:

PR #3356, opened 2016-02-23
Prevent AFImageDownloader crashing when passing invalid NSURLRequest (#3343)

xianwenx:
This references issues #3343, where an image download request can crash if it doesn't have a valid URL. Whether this should be handled by the framework or by the implementer is a design decision. It appears that NSAssert is not used for cases like this in the framework, so we are failing by returning a nil image download receipt instead. And given the rest of the behavior in the method, it is suggested the failure block should not be called (or a task created in the first place given invalid inputs). I hope this commit aligns with the solution the authors of AFNetworking would suggest for this scenario.

Feedback is greatly appreciated.

### DIFF
--- a/Tests/Tests/AFImageDownloaderTests.m
+++ b/Tests/Tests/AFImageDownloaderTests.m
@@ -64,6 +64,22 @@
     XCTAssertNil(self.downloader, @"Downloader should be nil");
 }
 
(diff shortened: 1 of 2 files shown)


kcharwood:
Thanks for the PR!

I've taken your approach, and refactored it to vend back a NSError in #3343. Can you take a look at that?

PR #3362, opened 2016-02-29
Improve AFAutoPurgingImageCache performance.

jasonhotsauce:
* The current implementation takes O(nlogn) time to add and purge image caches because we have to sort the dictionary. We can actually make it take O(n) time by using LRU Cache with Double LinkedList.


kcharwood:
Thanks @jasonhotsauce!

This is interesting. One thing I want to clarify. I think we currently only take the hit when adding when our memory capacity has been exceeded, not for _every_ image add, right?

Does it make sense to add any additional tests to this setup? Perhaps you could create a performance test demonstrating the improvement?

jasonhotsauce:
@kcharwood , Sounds good. I will create the performance test and compare these two different implementation. If the improvement isn't that significant as I expected, I think I would rather close this PR and using the current one. Thanks.

kcharwood:
That sounds good @jasonhotsauce. Looking forward to seeing the results.

jasonhotsauce:
@kcharwood I created the performance test and set the baseline as the current master implementation. You can see from the screen shot below that using LRU cache improved the performance 34%. 

screen shot 2016-03-14 at 11 18 42 pm

Thanks.

kcharwood:
Thanks for putting this together @jasonhotsauce.

For the time being, I'm going to leave this *_open_* as I'm not ready to pull in our own managed linked list into the library. So far, performance in this area has not been a major pressing issue, but I want to leave this open as an option for others to see and up vote if its causing them an issue.

:beers:

jasonhotsauce:
Sure. That sounds good. :beers:

kcharwood:
Given we still haven't seen much demand for this, I'm going to close this.

If we see pop up in the future, I'll be sure to reopen.

🍻

PR #3365, opened 2016-03-02
Added synchronous methods to AFHTTPSessionManager

yanivco:


kcharwood:
Thanks for the pr @yanivco!

This has been discussed a few times, and currently we have no plans to support synchronous networking within the core library. 

If you would like this functionality, I recommend maintaining it in your own branch.

:beers:

PR #3367, opened 2016-03-02
AFImageResponseSerializer tests

quellish:
Added a new test class specifically for AFImageResponseSerializer along with several tests.


kcharwood:
Thanks for the PR.

Could you please backmerge/rebase this with master so we can get tests passing again?

Thanks!

quellish:
Should be all set. All tests are passing for me except `testDomainReachabilityBlock` in the MacOS test target. That was failing before I started any work.

PR #3377, opened 2016-03-07
Fix carthage support 2.x

dulaccc:
Hey,

Because our app is heavily based on `NSOperation` usage, we are not able to upgrade to `3.0` for now. But the project wasn't compatible with Carthage, so I made this little fix (inspired by the compatibility work made on the 3.0 version)

Cheers


kaweepong2521:
good

kaweepong2521:
namdet k.

kaweepong2521:
You’re receiving notifications because you commented.namdet k.

kcharwood:
Thanks @dulaccc. However at this point, we are only taking critical security patches for 2.x.

The great thing about Carthage is you can just point it at your branch in your Cartfile, rather than `AFNetworking/AFNetworking`. For now, anyone looking for Carthage support for that version can look there.

:beers:

PR #3395, opened 2016-03-17
fix iamgeView no set

SummertimSadness:


kcharwood:
The documentation covers this use case:

> If a success block is specified, it is the responsibility of the block to set the image of the image view before returning. If no success block is specified, the default behavior of setting the image with `self.image = image` is applied.

:beers:

SummertimSadness:
other methods all set image, except this one,so i think
According to the normal logic,  it is the responsibility of the download method to set the image of the image view too，zhe block just want to get  image to do other things,maybe.

PR #3400, opened 2016-03-21
Patch for 1.3.x using Cocoapods - Reachability

mcfedr:
When using cocoapods `use_frameworks` its not possible to use AFNetworking reachability because of `#ifdef _SYSTEMCONFIGURATION_H`.

Although the framework includes SystemConfiguration in its PCH this isnt seen by the main project when the headers are included. 

As far as I can tell AFNetworking is compiled with the functionality, but we cannot see it from the project.

Simply removing these conditions fixes these problems.

P.S. I don't actually want this merged to master, but there is no 1.x branch. And I realise its probably 'unmaintained' but its depended on by some other libraries and upgrading is unfeasible at the moment, a 1.3.5 release would be really great!

### DIFF
--- a/AFNetworking.podspec
+++ b/AFNetworking.podspec
@@ -1,11 +1,11 @@
 Pod::Spec.new do |s|
   s.name     = 'AFNetworking'
-  s.version  = '1.3.2'
+  s.version  = '1.3.3'
   s.license  = 'MIT'
   s.summary  = 'A delightful iOS and OS X networking framework.'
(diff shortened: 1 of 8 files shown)


kcharwood:
Hi @mcfedr 

Thanks for the PR, but we have no plans to add or support any more work with AF 1.x.

I would recommending maintaining your own fork if there is specific functionality you need there.

:beers:

PR #3402, opened 2016-03-22
Adding extension point to customize User-Agent

dchohfi:
Adding category over `AFHTTPRequestSerializer` declaring `addUserAgentHeaderField` method that can be overridden by a subclass to customize values added to `User-Agent` header field.

I had to customize that value to add the specific device model on my requests, ended coping and pasting the logic contained on the `init` method.

What you guys think? Any changes or thoughts on that? Thank you.


kcharwood:
The User-Agent should be available to you through `- (nullable NSString *)valueForHTTPHeaderField:(NSString *)field;`

You can then edit it and set it to whatever you want with `- (void)setValue:(nullable NSString *)value
forHTTPHeaderField:(NSString *)field;`

For now, I'm going to decline this as I don't think its worth muddying the interface for this specific feature, given there is currently a way to set it.

:beers:

PR #3191, opened 2015-12-03
Changed SSL Pinning Error to be `NSURLErrorServerCertificateUntrusted`

kcharwood:
Cleaned up @0xced's work in #3169 and got the test to pass.

PR #3409, opened 2016-03-30
[WIP] AFSSLPinningModePublicKey allows request to proceed when server trust is invalid

brow:
When using an `AFSecurityPolicy` with pinning mode `AFSSLPinningModePublicKey`, when I make an HTTP request such that `-[AFSecurityPolicy evaluateServerTrust:forDomain:]` returns `NO`, then the request nevertheless proceeds and succeeds.

To demonstrate this, I've modified the test `-testInvalidServerTrustProducesCorrectError` in such a way that I believe it should still succeed, [but it fails](https://travis-ci.org/AFNetworking/AFNetworking/jobs/119416832).

This issue is apparently resolved by reverting #3191, but it's not clear to me that this is the correct thing to do.

### DIFF
--- a/Tests/Tests/AFHTTPSessionManagerTests.m
+++ b/Tests/Tests/AFHTTPSessionManagerTests.m
@@ -517,7 +517,7 @@
     NSData *googleCertificateData = [NSData dataWithContentsOfURL:googleCertificateURL];
     AFHTTPSessionManager *manager = [[AFHTTPSessionManager alloc] initWithBaseURL:[NSURL URLWithString:@"https://apple.com/"]];
     [manager setResponseSerializer:[AFHTTPResponseSerializer serializer]];
-    manager.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeCertificate withPinnedCertificates:[NSSet setWithObject:googleCertificateData]];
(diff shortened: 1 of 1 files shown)


kcharwood:
Hi @brow 

Could you be more specific? I'm slightly confused with your issue title, and your issue description. Your issue title indicates a requests succeeds when it should not, but your description seems to be the opposite: A request fails when it should succeed.

kcharwood:
I think I see what you are referring to now. If we return `NSURLSessionAuthChallengeRejectProtectionSpace` when we are pinning with public keys, that succeeds even though we told URL session to reject the protection space. Let me dig in on this, and thanks for bringing this to my attention.

:beers:

brow:
> I think I see what you are referring to now. If we return NSURLSessionAuthChallengeRejectProtectionSpace when we are pinning with public keys, that succeeds even though we told URL session to reject the protection space.

Yep, exactly. The test fails because it's expecting the request to fail with an error about server trust validation having failed, but it's actually failing with an error about the response content type because server trust validation erroneously succeeded. I could have made the test a bit clearer on this :see_no_evil: 

Also, the test as written will not go green if we simply revert #3191, because that change is what enabled the descriptive error.

Let me know if I can help. I was stumped after verifying that `-[AFSecurityPolicy evaluateServerTrust:forDomain:]` is correctly returning `NO`.

kcharwood:
Let's continue this in #3417 :beers:

PR #3417, opened 2016-03-31
Reverted `NSURLSessionAuthChallengeDisposition` to `NSURLSessionAuthChallengeCancelAuthenticationChallenge` for SSL Pinning

kcharwood:
As first discussed in #3409, there appear to be cases where returning `NSURLSessionAuthChallengeRejectProtectionSpace` does not prevent the request from succeeding in some cases.

A few of us have starting to dig in on this issue. @0xced has seen behavior where the existing code does work in some cases, leading us to believe there may be some issues here related to the TLS cache under the hood that we do not control. However, using `Cancel` instead does give us deterministic control.

The [original intent](https://github.com/AFNetworking/AFNetworking/pull/3169) of moving to `RejectProtectionSpace` was to provide a better error message to the developer to distinguish between a manual cancel and an SSL pinning failure.

I've created a simple playground demonstrating the issue:

[NSURLSessionAuthChallenge.playground.zip](https://github.com/AFNetworking/AFNetworking/files/197891/NSURLSessionAuthChallenge.playground.zip)
 
This PR reverts to using cancel, and adds an additional test for Public Key Pinning error validation.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -983,7 +983,7 @@ didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
                     disposition = NSURLSessionAuthChallengePerformDefaultHandling;
                 }
             } else {
-                disposition = NSURLSessionAuthChallengeRejectProtectionSpace;
+                disposition = NSURLSessionAuthChallengeCancelAuthenticationChallenge;
             }
         } else {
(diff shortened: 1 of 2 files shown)


cnoon:
This all looks good to me @kcharwood. 👍🏼

Would it also be wise to put the Apple Forums link in this PR as well for reference?

kcharwood:
Thanks @cnoon 

Here is a link to a Dev Forum I have opened:

https://forums.developer.apple.com/message/128153#128153

Issue #2520, opened 2015-01-15
Authentication block callbacks don't facilitate asynchronous handling


danielpunkass:
I am enjoying getting to know AFNetworking 2 and have generally been satisfied by most of the default decisions and behaviors of the library. The exceptions have all been easy to override and I can see your reasoning for defaulting to them.

One area that hung me up is in the way AFNetworking handles authentication challenges from NSURLSession, passing them along to the client asynchronously via registered callback blocks such as the "taskDidReceiveAuthenticationChallenge" block.

I saw in the discussion around issue #1335 that there was some surprise that NSURLSession's own delegate method uses an asynchronous callback block. The reasoning as far as I can understand it (and owing to the way I intend to use it) is that authentication could be delayed by the need to interact with the user e.g. to present a password prompt or other authentication flow. For this reason, the NSURLSession system seems designed to put a request on ice while the client obtains authorization from the user.

I think I could achieve what I need by taking advantage of the fact that the callback is on a non-main thread, blocking the thread while a prompt is displayed, and then returning the result synchronously to AFNetworking. But this seems a little shaky to me and also seems to be in defiance of the NSURLSession delegate expectation that the delegate should return promptly and call the callback when the credentials are ready. I don't know if the system puts the thread into any more useful of a sleeping state than I would waiting for my main thread UI to complete, but it does seem possible that it could have an effect on e.g. timeout timers or other machinery within NSURLSession.

I'm working around this by subclassing AFHTTPSessionManager and implementing my own delegate method for the authorization challenge, but I thought I would file an issue to get your take on it and get you thinking about whether it would be a good idea to facilitate NSURLSession's expected asynchronous callback mechanism in the longer term.

kcharwood:
Thanks for the write up here @danielpunkass, and I apologize for this sitting for so long with no interaction. I haven't seen anyone else talking about this feature, so its just been on the backburner.

I think your logic and reasoning does make sense. I'd love to see your implementation and see how it may fit back into AFNetworking. I'll do some thinking here as well.

:beers:

0xced:
I noticed the same issue while working on #3425. I’ll try to propose a solution that doesn’t require to subclass `AFURLSessionManager`.

danielpunkass:
[...]

PR #3425, opened 2016-04-04
Pinning failure better error

0xced:
Here’s another take at fixing #3165 which, unlike #3191, should not introduce a security flaw!

I will explain what I did and why in greater details later but I’m already submitting this pull request to see if all tests pass on Travis.

I’ll also create a better error with a full userInfo dictionary, so please don’t merge this yet.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -926,7 +926,7 @@ didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
  completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition disposition, NSURLCredential *credential))completionHandler
 {
     NSURLSessionAuthChallengeDisposition disposition = NSURLSessionAuthChallengePerformDefaultHandling;
-    __block NSURLCredential *credential = nil;
+    NSURLCredential *credential = nil;
 
     if (self.sessionDidReceiveAuthenticationChallenge) {
         disposition = self.sessionDidReceiveAuthenticationChallenge(session, challenge, &credential);
@@ -977,7 +977,7 @@ didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
  completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition disposition, NSURLCredential *credential))completionHandler
 {
     NSURLSessionAuthChallengeDisposition disposition = NSURLSessionAuthChallengePerformDefaultHandling;
-    __block NSURLCredential *credential = nil;
+    NSURLCredential *credential = nil;
 
     if (self.taskDidReceiveAuthenticationChallenge) {
         disposition = self.taskDidReceiveAuthenticationChallenge(session, task, challenge, &credential);
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -181,7 +181,7 @@ typedef void (^AFURLSessionTaskCompletionHandler)(NSURLResponse *response, id re
     }
 }
 
-static const void * const ServerTrustErrorKey = &ServerTrustErrorKey;
+static const void * const AuthenticationChallengeErrorKey = &AuthenticationChallengeErrorKey;
 
 static NSError * ServerTrustError(SecTrustRef serverTrust, NSURL *url)
 {
@@ -204,7 +204,7 @@ static NSError * ServerTrustError(SecTrustRef serverTrust, NSURL *url)
               task:(NSURLSessionTask *)task
 didCompleteWithError:(NSError *)error
 {
-    error = objc_getAssociatedObject(task, ServerTrustErrorKey) ?: error;
+    error = objc_getAssociatedObject(task, AuthenticationChallengeErrorKey) ?: error;
     __strong AFURLSessionManager *manager = self.manager;
 
     __block id responseObject = nil;
@@ -991,7 +991,7 @@ didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
             disposition = NSURLSessionAuthChallengeUseCredential;
             credential = [NSURLCredential credentialForTrust:challenge.protectionSpace.serverTrust];
         } else {
-            objc_setAssociatedObject(task, ServerTrustErrorKey, ServerTrustError(challenge.protectionSpace.serverTrust, task.currentRequest.URL), OBJC_ASSOCIATION_RETAIN);
+            objc_setAssociatedObject(task, AuthenticationChallengeErrorKey, ServerTrustError(challenge.protectionSpace.serverTrust, task.currentRequest.URL), OBJC_ASSOCIATION_RETAIN);
             disposition = NSURLSessionAuthChallengeCancelAuthenticationChallenge;
         }
     }
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -984,7 +984,9 @@ didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
 
     if (self.taskDidReceiveAuthenticationChallenge) {
         disposition = self.taskDidReceiveAuthenticationChallenge(session, task, challenge, &credential);
-    } else if ([challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodServerTrust]) {
+    }
+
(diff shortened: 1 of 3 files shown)


AFNetworking/AFURLSessionManager.m
```diff
@@ -253,6 +270,7 @@ - (void)URLSession:(__unused NSURLSession *)session
 {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wgnu"
+    error = objc_getAssociatedObject(task, ServerTrustErrorKey) ?: error;
```
cnoon: This is really slick @0xced! 🙌🏼

AFNetworking/AFURLSessionManager.m
```diff
@@ -1020,20 +1023,19 @@ - (void)URLSession:(NSURLSession *)session
...
-        if ([challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodServerTrust]) {
-            if ([self.securityPolicy evaluateServerTrust:challenge.protectionSpace.serverTrust forDomain:challenge.protectionSpace.host]) {
-                disposition = NSURLSessionAuthChallengeUseCredential;
-                credential = [NSURLCredential credentialForTrust:challenge.protectionSpace.serverTrust];
-            } else {
-                disposition = NSURLSessionAuthChallengeCancelAuthenticationChallenge;
-            }
+    }
+
+    if ([challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodServerTrust] && disposition == NSURLSessionAuthChallengePerformDefaultHandling) {
```
kcharwood: This is great @0xced!

My only question here is do we want to set an error if the user returns cancel with the `taskDidReceiveAuthenticationChallenge` block? Should we deprecate the existing block and add a parameter for setting an error, that we can then set if they return cancel?

0xced: I found a way without breaking compatibility at all. 😀 See ab8e777ee07c4d45deaeefc2bab39a0f9a09a59a.

AFNetworking/AFURLSessionManager.m
```diff
@@ -888,7 +906,7 @@ - (void)setTaskWillPerformHTTPRedirectionBlock:(NSURLRequest * (^)(NSURLSession
     self.taskWillPerformHTTPRedirection = block;
 }
 
-- (void)setTaskDidReceiveAuthenticationChallengeBlock:(NSURLSessionAuthChallengeDisposition (^)(NSURLSession *session, NSURLSessionTask *task, NSURLAuthenticationChallenge *challenge, NSURLCredential * __autoreleasing *credential))block {
+- (void)setTaskDidReceiveAuthenticationChallengeBlock:(NSURLSessionAuthChallengeDisposition (^)(NSURLSession *session, NSURLSessionTask *task, NSURLAuthenticationChallenge *challenge, id __autoreleasing *result))block {
```
kcharwood: Hmmmm.

I'm torn over this. I feel like this makes the API less readable now. My inclination is to create a new handler and deprecate the current one.

Thoughts?

0xced: Yep, that’s less readable, you're right. And if we're taking the deprecation route, we should probably take care of fixing #2520 at the same time. That would probably mean changing from 

```objc
NSURLSessionAuthChallengeDisposition (^)(NSURLSession *session, NSURLSessionTask *task, NSURLAuthenticationChallenge *challenge, NSURLCredential __autoreleasing *credential);
```
to
```objc
BOOL (^)(NSURLSession *session, NSURLSessionTask *task, NSURLAuthenticationChallenge *challenge, void (^)(NSURLSessionAuthChallengeDisposition disposition, NSURLCredential *credential));
```

Where the user would return `YES` from the block to indicates that the challenge was handled: responsibility to call `completionHandler(disposition, credential)` is upon the user. User would return `NO` to indicate that the challenge was not handled and AFNetworking's implementation (i.e. server trust evaluation) should be executed (responsibility to call `completionHandler(disposition, credential)` is upon AFNetworking)

Now the hardest part may be to find a new name for that new `taskDidReceiveAuthenticationChallenge` block.

kcharwood: Currently, if the user provides `taskDidReceiveAuthenticationChallenge`, they DO not also get the default pinning behavior. For backwards compatibility purposes, I would think we need to keep that behavior.

That means if they define the block, they are _fully_  responsible for it. If we were to modify that behavior, existing users who return a success would potentially still be susceptible to the call failing since pinning rules could then fail afterwards.

Essentially, I think if the block is defined, no other AF code should run to keep the current compatibility as close as possible to the existing behavior.

kcharwood: This may also mean we need the developer to potentially return the error they want to see reported?

0xced: > Currently, if the user provides `taskDidReceiveAuthenticationChallenge`, they DO not also get the default pinning behavior.

Exactly and I think this was a mistake, especially since it was not documented. I think it’s fair to fix that mistake and properly document it in the release notes.

> existing users who return a success would potentially still be susceptible to the call failing since pinning rules could then fail afterwards

I think that’s the right thing to do. Server trust failing when you pinned certificates is what you would expect IMO, regardless of if you set a `taskDidReceiveAuthenticationChallenge` or not. Again, the current documentation (3.1.0) does **not** state that having a `taskDidReceiveAuthenticationChallenge` bypasses the security policy.

0xced:
I tried something with 3e01f58c419fd573ecbe7b9eb8c9d71098f57c6a. The documentation is lacking for now.

kcharwood:

0xced:
I rebased, but this is stil missing documentation, so I wouldn’t merge yet. Also, I think there might be some overlap with #3616.

kcharwood:
Should I punt this from the 3.2.0 release train? Would like to release it this week sometime.

0xced:
I'm almost done writing the documentation. But it's not covered by tests at all for now. 😕

0xced:
I added documentation in 9310feb and I’m happy with it. It still needs some more tests though. I'd like to implement a custom authentication challenge handler that demonstrate how to use it for client certificates, but I think it’s [going to be tough](https://gist.github.com/0xced/9e16bebf82b37571fb6b3a24044523a3)!

dmiluski:
Any update on this pr? I’m a bit lost on how to integrate if I’m unable to differentiate cancel from pinning failure

n8gray:
Any chance this will get fixed? That "cancelled" error is a real pain.

jshier:
Unfortunately this PR is too complex to be merged with the amount of work it would take to validate the behavior.

0xced:
That's a shame, I invested so much time in it to have proper error handling. 🙁

jshier:
For future reference, the easier way to produce per-task security errors would be to drop the implementation of `URLSession:didReceiveChallenge:completionHandler:` altogether and rely on the `URLSession:task:didReceiveChallenge:completionHandler:` method. That way all auth challenges are handled on a per-task basis. If the `NSURLSession` level method is implemented at all it will receive all trust challenges.

jshier:
If you want you can reengineer the patch on top of current master and feel free to drop the backward compatibility concerns for the major release. If you do want to do this, you can build off of the `afnetworking4` branch that's up right now.

AFNetworking/AFURLSessionManager.m
```diff
@@ -181,12 +182,30 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
...
+static NSError * ServerTrustError(SecTrustRef serverTrust, NSURL *url)
+{
+    NSBundle *CFNetworkBundle = [NSBundle bundleWithIdentifier:@"com.apple.CFNetwork"];
+    NSString *defaultValue = @"The certificate for this server is invalid. You might be connecting to a server that is pretending to be “%@” which could put your confidential information at risk.";
+    NSString *descriptionFormat = NSLocalizedStringWithDefaultValue(@"Err-1202.w", nil, CFNetworkBundle, defaultValue, @"") ?: defaultValue;
+    NSString *localizedDescription = [descriptionFormat componentsSeparatedByString:@"%@"].count <= 2 ? [NSString localizedStringWithFormat:descriptionFormat, url.host] : descriptionFormat;
+    NSDictionary *userInfo = @{
+        NSURLErrorFailingURLErrorKey: url,
+        NSURLErrorFailingURLStringErrorKey: url.absoluteString,
+        NSURLErrorFailingURLPeerTrustErrorKey: (__bridge id)serverTrust,
```
Kaspik, line 196: This line is causing problems on new AFNetworking 4.0.0.

https://github.com/AFNetworking/AFNetworking/issues/4541

PR #2946, opened 2015-09-05
Add rss+xml as acceptable content type for AFXMLParserResponseSerializer

Donohue:
Signed-off-by: Brian Donohue <brian.donohue89@gmail.com>

PR #3436, opened 2016-04-12
为acceptableContentTypes添加两种常用类型

DXSmile:
为acceptableContentTypes添加两种常用类型  : @"text/plain", @"text/html"

Add two types of commonly used for acceptableContentTypes.
Reasons: we in the use of this framework, often not analysis from the server to the data type. Prompt content is the type is not supported. Which need us to write a class to inherit, manual for acceptableContentTypes add we need the type. And updated every frame, we all need to double check these types. This resulted in the a lot of unnecessary trouble. It is recommended that the author add we used type.

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -225,7 +225,7 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
         return nil;
     }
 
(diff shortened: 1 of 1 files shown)


0xced:
This has already been discussed in #3275, #3202, #3077, #2946, #2871, #2639 and #2447.

TLDR: Use [-[AFHTTPResponseSerializer setAcceptableContentTypes:]](http://cocoadocs.org/docsets/AFNetworking/3.1.0/Classes/AFHTTPResponseSerializer.html#//api/name/acceptableContentTypes)

PR #3441, opened 2016-04-14
extract the user agent generation from the requestSerialization init.

rapinto:
It allows easier customization into requestSerialization subclasses.


jshier:

PR #3472, opened 2016-04-28
Short-circuit unnecessary calls to SecTrustEvaluate() in pinning mode

Chm:
Assuming SecTrustEvaluate() was very slow, it's better to bypass certificate check before those were added to anchor.


kcharwood:
Could you create some performance tests demonstrating the difference?

jshier:
This change has been made separately, thanks.

PR #3476, opened 2016-04-28
Updates `AFHTTPSessionManager` documentation to reflect v3.x changes

### LINKED ISSUE #3412: Overriding dataTaskWithRequest:completionHandler:
Hey guys,

I am trying to override dataTaskWithRequest:completionHandler:. I am following your instructions found in the documentation for AFHTTPSessionManager where it is stated:

> To change the behavior of all data task operation construction, which is also used in the `GET` / `POST` / et al. convenience methods, override `dataTaskWithRequest:completionHandler:`.

My class subclasses AFHTTPSessionManager and in the implementation I have the dataTaskWithRequest:completionHandler: method. In its implementation I call [super  dataTaskWithRequest:completionHandler:].

Anyway, the problem is that dataTaskWithRequest:completionHandler: is never called...

Could you please check whether this is a bug? If not, please provide me some instructions.

Cheers!

ecaselles:
This closes #3412.

The documentation on `AFHTTPSessionManager`, *Methods to Override* section  states that...
> To change the behavior of all data task operation construction, which is also used in the `GET` / `POST` / et al. convenience methods, override `dataTaskWithRequest:completionHandler:` 

But that is no longer accurate, since these convenience methods now go through the new 
`dataTaskWithRequest:uploadProgress:downloadProgress:completionHandler:`. Therefore, overriding `dataTaskWithRequest:completionHandler:` will have no effect since it will never be called.

This PR updates the documentation to reflect the fact that the method to override is now `dataTaskWithRequest:uploadProgress:downloadProgress:completionHandler:`.

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.h
+++ b/AFNetworking/AFHTTPSessionManager.h
@@ -44,7 +44,7 @@
 
  ## Methods to Override
 
- To change the behavior of all data task operation construction, which is also used in the `GET` / `POST` / et al. convenience methods, override `dataTaskWithRequest:completionHandler:`.
+ To change the behavior of all data task operation construction, which is also used in the `GET` / `POST` / et al. convenience methods, override `dataTaskWithRequest:uploadProgress:downloadProgress:completionHandler:`.
 
  ## Serialization
 


kcharwood:
Thanks!

PR #3488, opened 2016-05-01
Updated character

gemmakbarlow:
Updated to a different unicode dash (`U+002D` rather than `U+2013`) so that text is parsed correctly by markdown (when converting the CocoaPods `Acknowledgements.markdown` to HTML). 

Prior to the adjustment, this was the result upon conversion to HTML:

markdown-issue-prior

With the adjustment, issue corrected: 

markdown-issue-after


kcharwood:
Thanks!

gemmakbarlow:
Welcome :)

PR #3490, opened 2016-05-03
Added deprecated attribute to  `dataTaskWithRequest:completionHandler:`

dchohfi:
It's dangers to leave `dataTaskWithRequest:completionHandler:` method without deprecated mark, as this method is now overloaded, and calling a different one, it breaks compatibility with previous overridden methods.


dchohfi:
@kcharwood what you think about that?

zhw511006:
which one is preferred? Thx!

dchohfi:
hey @zhw511006, you should be using the new one, as the method isn't called internally. The method is marked as deprecated to prevent subclasses from overriding it.

Issue #3498, opened 2016-05-09
Why AFNetworkReachabilityManager support IPv6 need to do conditional compile?


wtlucky:
I don't understand why there needs to check the OS_VERSION_MIN_REQUIRED. This mean only the iOS version over 9.0 and OSX version over 10.11 supported IPv6. Is this right?  #3174 

``` objc
+ (instancetype)manager
{
#if (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 90000) || (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 101100)
    struct sockaddr_in6 address;
    bzero(&address, sizeof(address));
    address.sin6_len = sizeof(address);
    address.sin6_family = AF_INET6;
#else
    struct sockaddr_in address;
    bzero(&address, sizeof(address));
    address.sin_len = sizeof(address);
    address.sin_family = AF_INET;
#endif
    return [self managerForAddress:&address];
}
```

balloonsys:
Maybe you can refer "Use System APIs to Synthesize IPv6 Addresses" section in following page -
[https://developer.apple.com/library/ios/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/UnderstandingandPreparingfortheIPv6Transition/UnderstandingandPreparingfortheIPv6Transition.html#//apple_ref/doc/uid/TP40010220-CH213-SW16](url)

kcharwood:
Take a look at this issue: https://github.com/Alamofire/Alamofire/issues/1228.

We need to make sure we are consistent in both places.

taylesworth:
Except AlamoFire uses a run-time check rather than a compile-time check, right?

kcharwood:
Yes, that rights, mainly due to some bugs with how iOS 8 work with IPv6

wtlucky:
Yes，I have found this bug both iOS 7 and iOS 8 works with IPv6. But why not use run-time check there? If I use the iOS 9 SDK compile the project and running it on iOS 7, the bug still exist.

kcharwood:
Right. I'm agreeing with you. We should probably migrate to a runtime check.

hei9gag:
[...]

PR #3501, opened 2016-05-11
Change compile-time check to runtime check.

wtlucky:
If use the iOS 9 SDK compile the project and running it on iOS 7 or iOS8. The ReachabilityManager will not run as expected.


cezheng:
I guess there's no need to use `sockaddr_in6` under any circumstances. See the reason I mentioned here: #3498 (comment)

PR #3540, opened 2016-05-26
Update AFNetworkReachabilityManager.m for IPv6

imrazor:
It's no needs to add sockaddr_in6 address.
If the build target is iOS9, in iOS8 device "setReachabilityStatusChangeBlock" not works in IPv4 network.
If the build target is lower than iOS9, in iOS8 device "setReachabilityStatusChangeBlock" not works in IPv6 network.
In iOS9 device "setReachabilityStatusChangeBlock" works good in both IPv4 & IPv6 network no mater which build target version is.
Base in Apple "Unofficial" review rules, they just test it in newest iOS version, so sockaddr_in is enough, add sockaddr_in6 leads to duplicated code, which is difficult to maintain.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -144,17 +144,11 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
 
 + (instancetype)manager
 {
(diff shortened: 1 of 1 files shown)


imrazor:
Our App has passed the Apple review by using this update

kcharwood:
I confirmed with Apple at WWDC that this code was not an issue for the IPV6 tests, and verified that in the networking lab on the WWDC IPV6 network.

We have seen issues where there were server side problems with IPv6 clients directly connecting, that required server side adjustments in order for some of Apple's tests to pass.

For now, I'm going to leave the code as is.

🍻

PR #3547, opened 2016-05-29
Fixed nullability errors

nrbrook:
Some tests are failing. I'm not sure if they failed before, I suspect they did. These changes should be reviewed carefully, some I wasn't too confident I was doing the right thing.


kcharwood:
We've only been converting the external .h files. Was there something specific that led you to modifying the .m files?

nrbrook:
I was including the source files directly in my project as I think when I first added AFNetworking there was no static library target. In my host project I have nullability warnings enabled.

Nullability is useful in ironing out bugs which could be caused by a nil object. Some of the checks I've added in this code will potentially mitigate such bugs.

jshier:

PR #3549, opened 2016-05-31
Use dispatch queue for reachability callbacks

SquaredTiki:
As opposed to using the main runloop. This uses the slightly newer `SCNetworkReachabilitySetDispatchQueue` function and may also prevent issues with the main run loop becoming blocked which I've seen occur myself and is mentioned here: http://stackoverflow.com/questions/34063166/ios-9-app-freeze-with-console-log-bksendhidevent.

Feedback on whether or not this is a good idea is very much appreciated 😊


kcharwood:
This would be a pretty big change for anyone with existing code right? People today assume main queue callback, and this would change that.

I'm not sure thats a change I would take, unless we dispatched to the main queue by default.

jshier:

PR #3565, opened 2016-06-11
Moved `[self didChangeValueForKey:@"currentState"]` into correct scope

chenxin0123:
I found '[self didChangeValueForKey:@"currentState"];' is outsite of the if{} scope,that confused me, maybe it should be insite.


kcharwood:
Thanks!

PR #3604, opened 2016-07-05
Fixed an issue where registering a UIProgessView to a task that was a…

Starscream27:
Noticed that when having a UIProgressView calling `setProgressWithUploadProgressOfTask:animated: `or `setProgressWithDownloadProgressOfTask:animated` of `UIProgressView+AFNetworking` with the task parameter object **having already completed** (status == NSURLSessionTaskStateCompleted), the app would crash when the task object is deallocated with the following error:

> CRASH: Fatal exception: An instance ***** of class ******* was deallocated while key value observers were still registered with it.


kcharwood:
Thanks!

Issue #3380, opened 2016-03-11
Failed Background Task Causes Crash (3.x)


Hackmodford:
I've noticed if my app is suspended while downloading in a background session, and the download fails (I think) I get this error message.

```
2016-03-10 22:29:05.172 Big Finish[516:130343] *** -[NSLock unlock]: lock (<NSLock: 0x13dd8ba70> 'com.alamofire.networking.session.manager.lock') unlocked when not locked
2016-03-10 22:29:05.173 Big Finish[516:130343] *** Break on _NSLockError() to debug.
2016-03-10 22:29:05.176 Big Finish[516:130343] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x13dddd800 of class __NSCFBackgroundDataTask was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x13f1b0570> (
<NSKeyValueObservance 0x13de4fd30: Observer: 0x13ddd65f0, Key path: countOfBytesReceived, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x13de4f0a0>
<NSKeyValueObservance 0x13f1aee60: Observer: 0x13ddd65f0, Key path: countOfBytesExpectedToReceive, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x13dde66a0>
<NSKeyValueObservance 0x13def1870: Observer: 0x13ddd65f0, Key path: countOfBytesSent, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x13f1af4a0>
<NSKeyValueObservance 0x13f1b0540: Observer: 0x13ddd65f0, Key path: countOfBytesExpectedToSend, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x13def28a0>
)

*** First throw call stack:
(0x182b5d900 0x1821cbf80 0x182b5d848 0x18349f5c0 0x182a45380 0x18260d8e8 0x18345881c 0x18350680c 0x1006d1bb0 0x1006de6c8 0x1006d58a0 0x1006e07e0 0x1006e04d8 0x1827c5470 0x1827c5020)
libc++abi.dylib: terminating with uncaught exception of type NSException`
```

I think this is caused by a failed download because Crashlytics always seems to log a failed download.

Here's an example:

[...]

PR #3607, opened 2016-07-06
Remove KVO of progress in favor of using the NSURLSession delegate APIs

coreyfloyd:
The impetus for this change was a bug related to:
https://github.com/AFNetworking/AFNetworking/issues/3380

Which points to an issue where AFURLSessionManager could get in a state where it does not inform the AFURLSessionManagerTaskDelegate to un-observe cancelled (or otherwise completed) tasks before they are deallocated.
While the above issue is specific to background tasks/sessions, we have observed it on sessions with default configurations as well.

Instead of debugging the KVO issues, I decided to replace the KVO code with in favor of the appropriate NSURLSession delegate methods.
Running all unit tests shows that this approach appears to be working.

The pull request that made this change (https://github.com/AFNetworking/AFNetworking/pull/3187) mentions that its goal was to remove the need for AFNetworking API consumers to use a KVO API for getting progress.
This makes sense, but I was unable to see any reason why KVO was used internally for getting progress information rather than the NSURLSession Delegate methods.
Because of this, It isn't clear to me what the historical reasons were/are for KVO being needed to satisfy the progress requirements of the AFNetworking API.

With any luck, KVO is no longer needed and we can eliminate this class of crashes by eliminating KVO.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -184,24 +184,6 @@ typedef void (^AFURLSessionTaskCompletionHandler)(NSURLResponse *response, id re
         }];
     }
 
-    [task addObserver:self
-           forKeyPath:NSStringFromSelector(@selector(countOfBytesReceived))
-              options:NSKeyValueObservingOptionNew
-              context:NULL];
-    [task addObserver:self
-           forKeyPath:NSStringFromSelector(@selector(countOfBytesExpectedToReceive))
-              options:NSKeyValueObservingOptionNew
-              context:NULL];
-
-    [task addObserver:self
-           forKeyPath:NSStringFromSelector(@selector(countOfBytesSent))
-              options:NSKeyValueObservingOptionNew
-              context:NULL];
-    [task addObserver:self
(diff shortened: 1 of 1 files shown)


coreyfloyd:
@kcharwood I'd like to get your thoughts on this PR since I saw you were the author of #3187 which added the KVO.

Is it possible to shed some light on why the delegate methods were removed and replaced with KVO? I assume that either the APIs changed or some of the APIs were unreliable, but the unit tests didn't reveal any issues.

Thanks for looking

reidmain:
What is the status of this PR? Could we get one of the AFNetworking contributors (cc: @kcharwood ) to look over this because I am experiencing crashes caused by this problem.

kcharwood:
Sorry @coreyfloyd for just now looking. Been a busy summer with lots of travel in Swift 3 migrations every two weeks...

I think I originally went the KVO route because of documentation I read somewhere, but I can't seem to find that at the moment...

Looking at the docs now, Apple clearly says to use these delegate callback methods, so I think this is a good approach. I'll pull it down and run the unit tests to verify 👍

coreyfloyd:
@kcharwood thanks for looking this over and getting this merged. Sorry I missed your comment earlier. I guess the tests passed for you locally?

PR #3642, opened 2016-08-11
change ipv6 condition in ReachabilityManager

dulingkang:
when run in more than iOS9 device, can not jump to the ipv6 branch in the master version. So I fix it use NSFoundationVersionNumber.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -144,7 +144,7 @@ static void AFNetworkReachabilityReleaseCallback(const void *info) {
 
 + (instancetype)manager
 {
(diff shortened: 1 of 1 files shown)


dulingkang:
this code want to execute ipv6 after iOS9, but it didn't

发自我的 iPhone

dulingkang:
```
    struct sockaddr_in6 address;
    bzero(&address, sizeof(address));
    address.sin6_len = sizeof(address);
    address.sin6_family = AF_INET6;
```
I think after iOS9 device, we want to execute above code, but actually, it didn't. So I currect the condition to above:
```
+    #if (TARGET_OS_IPHONE && NSFoundationVersionNumber > 1240) || (TARGET_OS_MAC && NSFoundationVersionNumber >= 1199)
```

ElfSundae:
The current condition is exactly correct, this PR can be closed.
ping @jshier

PR #3653, opened 2016-08-19
Handle Error Pointers according to Cocoa Convention

tclementdev:
This is a Cocoa convention.

### DIFF
--- a/AFNetworking.xcodeproj/project.pbxproj
+++ b/AFNetworking.xcodeproj/project.pbxproj
@@ -197,6 +197,9 @@
 		5F4323DD1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer in Resources */ = {isa = PBXBuildFile; fileRef = 5F4323DC1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer */; };
 		5F4323DE1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer in Resources */ = {isa = PBXBuildFile; fileRef = 5F4323DC1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer */; };
 		5F4323DF1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer in Resources */ = {isa = PBXBuildFile; fileRef = 5F4323DC1BF63CCC003B8749 /* GeoTrust_Global_CA_Root.cer */; };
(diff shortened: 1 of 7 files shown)


kcharwood:
Hi @tclementdev! Thanks for the PR.

Apple's example docs in [Dealing with Errors](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/ErrorHandling/ErrorHandling.html) uses the standard we are currently using in this library, so I'm going to stick with that for now.

```Objective-C
    NSError *anyError;
    BOOL success = [receivedData writeToURL:someLocalFileURL
                                    options:0
                                      error:&anyError];
    if (!success) {
        NSLog(@"Write failed with error: %@", anyError);
        // present error to user
    }
```

🍻

tclementdev:
I think you missed the point. The return value of the method call needs to be tested before looking at the error. This is what is done in the sample code from Apple you provided and this is not what is done in the AFNetworking code.

tclementdev:
@kcharwood

0xced:
I don’t understand why this PR was closed, @tclementdev point is perfectly valid.

From Cocoa's [Error Handling Programming Guide](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html#//apple_ref/doc/uid/TP40001806-CH204-SW1)
> **Important**: Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning `nil` or `NO`, you should always check that the return value is `nil` or `NO` before attempting to do anything with the `NSError` object.

kcharwood:
Sorry misread the original intent! Reopened here.

Has a full audit of the codebase been done for this particular issue?

tclementdev:
Thank you for re-opening this PR. I just pushed some additionnal fixes concerning the same issue in other system API calls.

I could not find any other system API misuse, however it should be noted that the AFNetworking APIs themselves seem to be quite oblivious of this convention, with code seemingly returning non-nil objects while an error occurred internally. There is probably some amount of work to audit and modify the AFNetworking code itself to clarify and address this (this can be done in a separate PR).

kcharwood:

tclementdev:
Done.

kcharwood:
You can see the codecov diff [here](https://codecov.io/gh/AFNetworking/AFNetworking/compare/e337471d80a554c8f3e8b0a53689c3ed75cb47c9...2201d2cd555f494ed25f96c0d0d940bc018b8933).

Verifying those nil's should help prevent regressions in the future. After that, should be good to merge!

tclementdev:
Am I good?

kcharwood:
🍻 thanks!

PR #3668, opened 2016-09-01
change bool value usage

zhaoyuyu:
bool value should be used in this way


Huang-Libo:
make no different..

kcharwood:
Thanks for the PR @zhaoyuyu, but I'm going to decline this for now to maintain consistency.

🍻

PR #3675, opened 2016-09-06
Removing NULL objects within an Array

vsubrahmanian:
This method currently removes keys for NULL values only within a dictionary.
Proposing here to remove NULL objects from Array as well.

If this goes against the function name _AFJSONObjectByRemovingKeysWithNullValues_ and the flag _removesKeysWithNullValues_ would you recommend changing them (to _AFJSONObjectByRemovingNullValues_ and _removesNullValues_ respectively) or add another Flag?

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.m
+++ b/AFNetworking/AFURLResponseSerialization.m
@@ -64,7 +64,9 @@ static id AFJSONObjectByRemovingKeysWithNullValues(id JSONObject, NSJSONReadingO
     if ([JSONObject isKindOfClass:[NSArray class]]) {
         NSMutableArray *mutableArray = [NSMutableArray arrayWithCapacity:[(NSArray *)JSONObject count]];
         for (id value in (NSArray *)JSONObject) {
-            [mutableArray addObject:AFJSONObjectByRemovingKeysWithNullValues(value, readingOptions)];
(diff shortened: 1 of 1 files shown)


0xced:
I’m not sure that removing nulls from arrays is a good idea. Removing null from dictionaries is fine because `dictionary[@"key"]` returns nil when the key does not exist and nil is much easier to deal with than NSNull.
Removing nulls from an array is dangerous because it will change the size of the array and you will get bad surprises when accessing an array by index. For example, given `@[ [NSNull null], @"one", @"two" ]` and null removal, accessing `array[1]` will return `@"two"`, which is probably not what you expected.

vsubrahmanian:
@0xced Completely agree. 
I was thinking about the case when we are trying to send messages to objects in the array. If there is a probability of values turning up as _null_, we will need to add _null_ checks all over before doing so.
However, the case you talked about, is indeed a strong reason for why we shouldn't probably do this. Also I infer that it will complicate any debugging if we don't see the complete array.
Thank you! Closing this. 
[going back to adding _null_ checks in my project :P If you have suggestions for the case I talked about, let me know! Thanks again :)]

kcharwood:
Thanks @0xced. I agree with you on that point.

🍻

PR #3682, opened 2016-09-13
Loading an image from Amazon S3 server

umakanta:
Loading an image from Amazon S3 server


kcharwood:
Thanks @umakanta. A `binary/octect-stream` is not always an image, so it does not make sense to apply this as a default. You should add this to your response serialization in your codebase using the `acceptableContentTypes` property on your response serializer. 

🍻

PR #3687, opened 2016-09-19
AFHTTPSessionManager now throws exception if SSL pinning mode is set for non https sessions

0xced:
### Before this commit
Setting a security policy configured with `AFSSLPinningModeCertificate` or `AFSSLPinningModePublicKey` on a AFHTTPSessionManager instance configured with an insecure `http` base URL was valid. Requests made with this manager would always succeed since the `-[AFURLSessionManager URLSession:didReceiveChallenge:completionHandler:]` would never be called and thus the security policy would never be evaluated.

### After this commit
Setting a security policy configured with `AFSSLPinningModeCertificate` or `AFSSLPinningModePublicKey` on a AFHTTPSessionManager instance configured with an insecure `http` base URL will throw an exception. This will force the manager to be configured with a secure `https` URL.

Note that properly configuring App Transport Security (ATS) would also solve this issue since insecure connections would fail anyway, but this is a *belt and suspenders* solution.

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.m
+++ b/AFNetworking/AFHTTPSessionManager.m
@@ -98,6 +98,23 @@
     [super setResponseSerializer:responseSerializer];
 }
 
+@dynamic securityPolicy;
+
+- (void)setSecurityPolicy:(AFSecurityPolicy *)securityPolicy {
+    if (securityPolicy.SSLPinningMode != AFSSLPinningModeNone && ![self.baseURL.scheme isEqualToString:@"https"]) {
+        NSString *pinningMode = @"Unknown Pinning Mode";
+        switch (securityPolicy.SSLPinningMode) {
+            case AFSSLPinningModeNone:        pinningMode = @"AFSSLPinningModeNone"; break;
+            case AFSSLPinningModeCertificate: pinningMode = @"AFSSLPinningModeCertificate"; break;
+            case AFSSLPinningModePublicKey:   pinningMode = @"AFSSLPinningModePublicKey"; break;
+        }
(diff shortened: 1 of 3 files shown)


AFNetworking/AFHTTPSessionManager.m
```diff
@@ -98,6 +98,22 @@ - (void)setResponseSerializer:(AFHTTPResponseSerializer <AFURLResponseSerializat
     [super setResponseSerializer:responseSerializer];
 }
 
+- (void)setSecurityPolicy:(AFSecurityPolicy *)securityPolicy {
```
kcharwood: Should we add a note to the documentation declaring this behavior?

kcharwood: I think that would be a good idea, since that behavior only affects `AFHTTPSessionManager`

0xced: OK, I just rebased on master and force-pushed with documentation.

0xced:
After [much struggling](https://twitter.com/0xced/status/778862550683357184), all test finally pass!

kcharwood:
@0xced awesome!

I think it would make sense to cherry-pick the tests stability into its on PR, so we can have a direct reference to that and it doesn't get lost in this changeset. Would you be able to do that? If not, I can take a stab at it.

0xced:
Yes, I will do it.

kcharwood:

0xced:
I rebased, but I think you should be able to [rebase and merge](https://github.com/blog/2243-rebase-and-merge-pull-requests) from the GitHub interface now.

kcharwood:
I wanted to see the green check before merging!

0xced:
I realized right after posting. 😄

PR #3694, opened 2016-09-22
Fix reachability tests

0xced:
Hopefully these changes will fix the domain reachability tests.


kcharwood:
Now that we have passing tests, I'm not sure we need this. Thoughts?

PR #3718, opened 2016-09-30
Fixed potential KVO crasher for URL Session Task delegates

0xced:
Please see both commit comments for a full explanation.


kcharwood:
This is great work @0xced! I was actually going down the same rabbit hole last week before my vacation, so I agree with all of your assessment here!

I'm verifying these changes locally as well right now. Will merge soon assuming I find no other issues.

🍻

PR #3726, opened 2016-10-05
Disable Background URL Session Tests for Xcode 8

kcharwood:
I've discovered that background url sessions are not very reliable in test targets in Xcode 8. I've disabled them for now, and filed rdar://28588583


kcharwood:
Thanks to @jshier for `FASTLANE_EXPLICIT_OPEN_SIMULATOR=1`

PR #3733, opened 2016-10-09
Fixed issue where UIWebView extension did not preserve all of the request information

### LINKED ISSUE #3724: Parameter set to nil on UIWebView+AFNetworking.m
I just upgrade my AFNetworking version from 2.5.4 to 3.1.0, and after test some feature on my app, my UIWebView doesn't work with a 401 error (Authentication Error). I have passing the Authentication header to the request, but then after checking the lib's code, I found this:

image

The parameters that supposed to be where the authentication header is, is set to nil. When I tried to debug the value of request, the header does exists, but since the code set the parameters to nil, then it return 401 error.

Any idea? Thanks.

skyline75489:
This should fix #3724 and any header-related issue in the future.

After this changes, the entire `request`, instead of only the `URL`, will be used to construct the `dataTask`.

Test case is added.

### DIFF
--- a/Tests/Tests/AFUIWebViewTests.m
+++ b/Tests/Tests/AFUIWebViewTests.m
@@ -24,6 +24,7 @@
 #import "UIWebView+AFNetworking.h"
 
 @interface AFUIWebViewTests : AFTestCase
+
 @property (nonatomic, strong) UIWebView *webView;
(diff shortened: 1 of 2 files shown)


kcharwood:
🍻 thanks!

PR #3747, opened 2016-10-11
Add test method to wait for common timeout without handler

skyline75489:
Since literally every call to `waitForExpectationsWithCommonTimeoutUsingHandler` is passing a `nil` parameter, I think there should be a convenient method for this. 

This simplify the waiting method when doing async test. I have a feeling that people would love this.


kcharwood:
Thanks! 🍻

PR #3748, opened 2016-10-11
Enable custom httpbin URL with HTTPBIN_BASE_URL environment variable

0xced:
Editing the scheme and checking the `HTTPBIN_BASE_URL` box enables running the test suites on a default local installation of httpbin.


kcharwood:
Would it make sense to install that during the CI script to speed up the test suite? I'm assuming the test suite runs faster?

0xced:
No, because it's running on http instead of https and thus broking some tests. I tried to run it on https  some time ago but failed. I’m not sure it would even be possible to get a valid certificate for 127.0.0.1 without a domain name.

kcharwood:
Ah good point. Didn't think about that.

PR #3751, opened 2016-10-13
Deprecate the unused stringEncoding property of AFHTTPResponseSerializer

0xced:
The `stringEncoding` property is never used. It was probably inroduced by error in 9a40c0dc0f27e9a547fb173a0b1b5a06072e91f9. It was probably a copy/paste error as the same property also exists on `AFHTTPRequestSerializer` where it’s actually used.

### DIFF
--- a/AFNetworking/AFURLResponseSerialization.h
+++ b/AFNetworking/AFURLResponseSerialization.h
@@ -57,10 +57,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init;
 
-/**
- The string encoding used to serialize data received from the server, when no string encoding is specified by the response. `NSUTF8StringEncoding` by default.
- */
-@property (nonatomic, assign) NSStringEncoding stringEncoding;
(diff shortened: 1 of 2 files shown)


skyline75489:
Is this sort of breaking backward compatibility ? If stick strictly to [semver](http://semver.org/), this should probably be in 4.0.0.

0xced:
I just marked the property as deprecated, so this is not breaking backward compatibility. Removing it would break compatibility and warrant a 4.0.0 version.

But even when introducing breaking changes, it’s always a good idea to deprecate and explain with a comment what the replacement is.

skyline75489:
Gotta it. Thanks for the explaination.

tcamin:
This looks like a major breaking change.

Before these changes the value of the property was set to `NSUTF8StringEncoding` = 4 in `AFURLResponseSerialization`'s initializer, now it's defaulted to 0 which is actually an invalid string encoding.

If you try to encode/decode using 0 (e.g. `[@"test" dataUsingEncoding:0]`) you'll be prompted with the following warning:
```Incorrect NSStringEncoding value 0x0000 detected. Assuming NSASCIIStringEncoding. Will stop this compatibility mapping behavior in the near future.``` 

So in the new implementation stringEncoding will be actually `NSASCIIStringEncoding` and not `NSUTF8StringEncoding` as it was before.

PR #3752, opened 2016-10-14
Search for default pinned certificates in the main bundle

### LINKED ISSUE #3575: Can't pinned certificates from bundle
defaultPinnedCertificates loaded certificates from [NSBundle bundleForClass:[self class]], but when AFNetworking imported like framework via CocoaPods, then bundle has bundle  has "*.app/Frameworks/AFNetworking.framework", but needs to use main app bundle.

0xced:
Using `[NSBundle bundleForClass:[self class]]` is dangerous because the code will behave differently depending on how AFNetworking is integrated.
- If it is integrated as a static library (for example using CocoaPods), certificates will be searched in the main bundle.
- If it is integrated as a framework (for example using Carthage), certificates will be searched in the AFNetworking framework.

Even though this behavior is documented, this is dangerous. Using the main bundle makes the behavior deterministic.

Fixes #3575

### DIFF
--- a/AFNetworking/AFSecurityPolicy.h
+++ b/AFNetworking/AFSecurityPolicy.h
@@ -45,10 +45,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  The certificates used to evaluate server trust according to the SSL pinning mode. 
-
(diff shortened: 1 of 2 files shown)


jshier:
Old, but still valid, and this change matches Alamofire's behavior.

Jackylee1992:
👍

PR #3762, opened 2016-10-19
Use `strongSelf` local variable

crazytonyli:
This line is added in [PR #3325](https://github.com/AFNetworking/AFNetworking/pull/3325/files#diff-9d9536229ce0eada1e165adb747fc722R232), I reckon it's supposed to use `strongSelf` rather than `self`.


tewha:
Unless this change has been made outside of this merge, this looks fairly important to me.

crazytonyli:
Any chance we can look into whether this is a proper fix or not? Top 1 crash in my app(and also https://github.com/AFNetworking/AFNetworking/issues/3876) might be caused by this.

3lvis:
@jshier Any chance that you could look into this?

stephenyao:
Any chance this could get merged, as we are experiencing many crashes in our app due to this

nikfil89:
This commit is implemented in newest 3.2.1 version release but the problem referenced in #3876 still persists.

jshier:
This has been fixed.

PR #3765, opened 2016-10-21
Check data validation if nil not only length before parse response Object from  NSJSONSerialization

StevenUpForever:
When using AFHTTPSessionManager, and response data is nil, it could get
data.length == 0, but when parse it in NSJSONSerialization, it will
crash the method


AFNetworking/AFURLResponseSerialization.m
```diff
@@ -244,7 +244,7 @@ - (id)responseObjectForResponse:(NSURLResponse *)response
     // See https://github.com/rails/rails/issues/1742
     BOOL isSpace = [data isEqualToData:[NSData dataWithBytes:" " length:1]];
     
-    if (data.length == 0 || isSpace) {
+    if (!data || data.length == 0 || isSpace) {
```
0xced: The condition you added (`!data`) changes absolutely nothing to the existing code. If data is nil, the `data.length == 0` condition will evaluate to true.

PR #3792, opened 2016-11-19
Correct nullability

nrbrook:
Xcode static analysis reveals -uploadTaskWithRequest:fromFile:progress:completionHandler: can return nil and should be marked as nullable.


jshier:

Issue #2970, opened 2015-09-13
AFNetworkReachabilityManager doesn't post which instance of itself is posting the notification.


donileo:
Hey guys is there any reason AFNetworkReachabilityManager doesn't post the sending object  in AFNetworkingReachabilityDidChangeNotification? Due to it not doing this theres no way to easily differentiate between the AFNetworking reachability manager posting and any other instances of AFNetworkReachabilityManager that I may instantiate for my own classes.

Also I am seeing weird behaviors such as multiple notifications coming when only should.

mobitar:
As an alternative, you might be able to try using the block callback instead of notifications for reachability changes:

On `AFNetworkReachabilityManager`:
```
- (void)setReachabilityStatusChangeBlock:(nullable void (^)(AFNetworkReachabilityStatus status))block;```

phoney:
I don't see any reason why self isn't passed for the object when that notification is posted. 

Having said that, I've always just used the singleton shared reachability manager in my code. I'm not sure what use case requires more than one instance.

PR #3811, opened 2016-12-12
Add sender object when posting AFNetworkingReachabilityDidChangeNotification

idcuesta:
## Description
This PR attempts to fix issue #2970 

When an app handles several `AFReachabilityManager` and uses `AFNetworkingReachabilityDidChangeNotification` at the same time, it is not possible to identify what manager is posting the notifications.

With proposed changes, object that act as sender of notifications is the underlying `SCNetworkReachabilityRef` of each manager (I have chosen this object because of practical reasons, as it was already available from callers of method `AFPostReachabilityStatusChange`). This object has been exposed in header file as readonly so it can be used to listen for specific notifications.


jshier:

PR #3813, opened 2016-12-14
Fixed a warning when linking from an app extension

### LINKED ISSUE #3398: "Linking against dylib not safe for..." when framework is installed through Carthage
Hello guys,

Xcode gives me the following warning for my watchOS extension when AFNetworking is added to the project through Carthage:

> (null): Linking against dylib not safe for use in application extensions: /Users/ivankonov/Documents/work/ios-fan-app-5/Carthage/Build/watchOS/AFNetworking.framework/AFNetworking

This is the script for the iOS target:

> $(SRCROOT)/Carthage/Build/iOS/AFNetworking.framework

and this is the script for the wathcOS target:

> $(SRCROOT)/Carthage/Build/watchOS/AFNetworking.framework

I have read some suggestions to disable the "Require Only App-Extension-Safe API" option but it does not seem like the best idea to me.

Thank you in advance!

vhbit:
When an app extension was using the framework built by `carthage` there
was a warning "linking against a dylib which is not safe for use in
application extensions".

This build setting makes linker much happier and mitigates the warning.

Fixes #3398 on iOS

### DIFF
--- a/AFNetworking.xcodeproj/project.pbxproj
+++ b/AFNetworking.xcodeproj/project.pbxproj
@@ -1370,6 +1370,7 @@
 		299522421BBF104D00859F49 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				BITCODE_GENERATION_MODE = marker;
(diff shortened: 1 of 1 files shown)


jshier:

PR #3840, opened 2017-01-05
Unwrap license text

vtourraine:
This pull request unwraps the license text.

This is purely a formal change, and I hope not too controversial. In some contexts we might want to display the license, for instance in a `UITextView`, where having artificial line breaks can become an issue.


jshier:

PR #3871, opened 2017-02-09
AFJSONObjectByRemovingKeysWithNullValues use less memory

guidev:
There's no need to create a mutable dictionary if it is already. 
This saves a lot of memory especially with big responses.


jshier:

PR #3872, opened 2017-02-10
fix AssociatedObject bug

welcommand:
example  UIImageView (AFNetworking)
 in "sharedImageDownloader:"  self is [UIImageView class].
than i have a class like this     UserImageView : UIImageView ,  in "sharedImageDownloader:"  self is [UserImageView class].
[UIImageView class] != [UserImageView class]
so  call [UserImageView sharedImageDownloader] , the objc_getAssociatedObject(self, @selector(sharedImageDownloader)) returns value is always nil.


tewha:
Nice find.

PR #3891, opened 2017-03-09
AFURLRequestSerialization: use dispatch_barrier_sync for header changes

### LINKED ISSUE #3636: Crashes in [AFHTTPRequestSerializer setValue:forHTTPHeaderField:]
I have started experiencing lots of crashes in this method: `[AFHTTPRequestSerializer setValue:forHTTPHeaderField:]`  after my latest app update.

It's both SIGSEGV and SIGABRT crashes but most of them have reasoning like this:

`*** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[<__NSCFConstantString 0x79e7b8> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key User-Agent.'`

or

`- [AFHTTPRequestSerializer setValue:forHTTPHeaderField:]
*** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[<AFJSONResponseSerializer 0x17605460> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key Accept-Language.'`

`AFURLRequestSerialization.m, line 316`

The name of the key varies and even the name of the class that was tried to `setValue` on.

My latest update included current AFNetworking `master` branch and I see there have been some changes in the code which crashes. Changes are in this commit: https://github.com/AFNetworking/AFNetworking/commit/2a53b2c3f49275352677dfd46a35b0287ae01458 I think it could be related.

I have lots of stacktraces for this crash and can update more info if needed.

Knut

tmm1:
This fixes a rare crash in the async block, which I believe occurs when the block was invoked async _after_ the object had already been deallocated.

fixes #3636

cc #3526 @aopod @alexbird

### DIFF
--- a/AFNetworking/AFURLRequestSerialization.m
+++ b/AFNetworking/AFURLRequestSerialization.m
@@ -312,7 +312,7 @@ static void *AFHTTPRequestSerializerObserverContext = &AFHTTPRequestSerializerOb
 - (void)setValue:(NSString *)value
 forHTTPHeaderField:(NSString *)field
 {
-    dispatch_barrier_async(self.requestHeaderModificationQueue, ^{
+    dispatch_barrier_sync(self.requestHeaderModificationQueue, ^{
         [self.mutableHTTPRequestHeaders setValue:value forKey:field];
     });
 }


tewha:
Doesn't `dispatch_barrier_async` strongly capture variables used in blocks?

tmm1:
Good point. So my theory as to why this fixes the crash is incorrect, however it does appear to be fixed.

tewha:
Probably related to atomicity. Interesting. I have no theories. :)

aopod:
It's weird. Can't explain it for now. I'm trying to figure it out. Can you reproduce the problem?

SandyChapman:
@tewha : I'm hitting this bug as well. If I had to guess, it's because of the following:

```
@property (readwrite, nonatomic, strong) NSMutableDictionary *mutableHTTPRequestHeaders;
@property (readwrite, nonatomic, strong) dispatch_queue_t requestHeaderModificationQueue;
```

```
self.requestHeaderModificationQueue = dispatch_queue_create("requestHeaderModificationQueue", DISPATCH_QUEUE_CONCURRENT);
```

```
dispatch_barrier_async(self.requestHeaderModificationQueue, ^{
    [self.mutableHTTPRequestHeaders setValue:value forKey:field];
});
```
So we have a concurrent queue that is writing to a `nonatomic` `NSMutableDictionary`. Looks like a threading issue since we can theoretically have two threads writing to this dictionary at the same time.

I'd suggest switching to a serial dispatch queue here instead. I'll open a PR shortly.

jshier:

tmm1:
@jshier The code in this area has not changed as far as I can tell, and the patch should still apply. This change stops many crashes we've seen in our app.

Do you want me to create a new PR with the same commit?

jshier:
@tmm1 Please do. Additionally, it seems like the same change should be made to `clearAuthorizationHeader`, no?

tmm1:
@jshier A new version of the PR has been created as requested: #4474. Would really appreciate if you could merge that as this fixes many crashes seen in the wild.

PR #3922, opened 2017-04-11
UIImageView and UIButton setImageWithURL: internally uses requestSerializer

lukaskukacka:
### Overview
`UIImageView` and `UIButton` both have categories with method `-setImageWithURL:placeholderImage:` (and shortcut `setImageWithURL:`).This method is building new `NSMutableURLRequest` and manually setting `Accept: image/*` header.

### Problems
1. It is impossible to customise HTTP headers and other request properties (e.g. cache policy etc.) globally for all images downloaded using the category
    - Only solution is to implement custom category always building custom request
2. All Image requests done using this convenience method have only basic `Accept: image/*`
3. Requests done using the method ignore even the `AFNetworking`'s basic default HTTP headers like `User-Agent`

### Suggestion / solution
IMHO the request done using the convenience method should be using request serializer of session manager used for this request (`[class sharedImageDownloader].sessionManager.requestSerializer`).
That way it is possible have global configuration for all image requests done using this method.

This change allows configuration like this:
```
AFHTTPRequestSerializer <AFURLRequestSerialization> * requestSerializer = [[UIImageView class] sharedImageDownloader].sessionManager.requestSerializer;
[requestSerializer setValue:customHeader forHTTPHeaderField:@"MyCustomValue"];
requestSerializer.cachePolicy = NSURLRequestReloadIgnoringCacheData;
```
(Please note that this is just an example. Turning off image cache completely might not very desirable 😃 )

### Compatibility
Without explicitly opting in by customising request serializer, the solution should work just as before the change. The same default `Accept` header is set, just on different place (removed duplication).

---

### Feedback welcome
Please note I opened this PR mainly as suggestion / base for discussion. I'd like to hear your opinion if my thinking about it is right.

### DIFF
--- a/UIKit+AFNetworking/AFImageDownloader.m
+++ b/UIKit+AFNetworking/AFImageDownloader.m
@@ -145,6 +145,7 @@
     NSURLSessionConfiguration *defaultConfiguration = [self.class defaultURLSessionConfiguration];
     AFHTTPSessionManager *sessionManager = [[AFHTTPSessionManager alloc] initWithSessionConfiguration:defaultConfiguration];
     sessionManager.responseSerializer = [AFImageResponseSerializer serializer];
+    [sessionManager.requestSerializer setValue:@"image/*" forHTTPHeaderField:@"Accept"];
 
     return [self initWithSessionManager:sessionManager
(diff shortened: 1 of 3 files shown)


jshier:

PR #3932, opened 2017-04-12
Image Request cancel bug.

RbBtSn0w:
About UIImageView+AFNetworking will cancelImageDownloadTask.
cancelTaskForImageDownloadReceipt is working on GCD, when we cancel the request at main thread. The first work at  "self.af_activeImageDownloadReceipt = nil;" , when CANCEL Block , if ([strongSelf.af_activeImageDownloadReceipt.receiptID isEqual:downloadID]) alway is false.

### DIFF
--- a/UIKit+AFNetworking/UIImageView+AFNetworking.m
+++ b/UIKit+AFNetworking/UIImageView+AFNetworking.m
@@ -140,7 +140,6 @@
 - (void)cancelImageDownloadTask {
     if (self.af_activeImageDownloadReceipt != nil) {
         [[self.class sharedImageDownloader] cancelTaskForImageDownloadReceipt:self.af_activeImageDownloadReceipt];
-        [self clearActiveDownloadInformation];
      }
 }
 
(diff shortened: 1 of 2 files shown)


UIKit+AFNetworking/UIImageView+AFNetworking.m
```diff
@@ -140,7 +140,6 @@ - (void)setImageWithURLRequest:(NSURLRequest *)urlRequest
 - (void)cancelImageDownloadTask {
     if (self.af_activeImageDownloadReceipt != nil) {
         [[self.class sharedImageDownloader] cancelTaskForImageDownloadReceipt:self.af_activeImageDownloadReceipt];
-        [self clearActiveDownloadInformation];
```
RbBtSn0w: Need remove this.

RbBtSn0w: At line 128. 
if ([strongSelf.af_activeImageDownloadReceipt.receiptID isEqual:downloadID])  
always fail.

PR #3934, opened 2017-04-16
Redirect response should return nullable NSURLRequest

ReDetection:
Return value from `- (NSURLRequest *)connection:(NSURLConnection *)connection willSendRequest:(NSURLRequest *)request redirectResponse:(NSURLResponse *)response;` 
> The actual URL request to use in light of the redirection response. The delegate may return request unmodified to allow the redirect, return a new request, or *return nil to reject the redirect and continue processing the connection*.

So in the header you have `ASSUME_NONNULL` block which breaks that particular line, so I fixed it.


ReDetection:
Travis is failing because buildscript installs always latest version of cocoapods, which is now incompatible with 2.x branch

jshier:

PR #3936, opened 2017-04-18
Fix imageRequestCancel bug.

RbBtSn0w:
About UIImageView+AFNetworking will cancelImageDownloadTask.
cancelTaskForImageDownloadReceipt is working on GCD, when we cancel the
request at main thread. The first work at
"self.af_activeImageDownloadReceipt = nil;" , when CANCEL Block , if
([strongSelf.af_activeImageDownloadReceipt.receiptID
isEqual:downloadID]) alway is false.


Kaspik:
👍 for this

jshier:

PR #3944, opened 2017-05-08
Fix urlquery

coppynight:
if request url ends up with "?",  requestBySerializingRequest method can't work perfectly. URL.query return "" instead of nil in this case.


jshier:
Thanks for the PR! However, due to its age it will not be merged. However, if the issue is still occurring feel free to open a new PR.

PR #3951, opened 2017-05-14
The copy should be strong.

wanshanhu79:


jshier:
This is incorrect, mutable values should always be `copy`.

PR #3954, opened 2017-05-23
Resolve memory leaks

hongcaiyu:
The AFURLSessionManager can not release.Because it references the session cycle。So I call the - (void)finishTasksAndInvalidate when mutableTaskDelegatesKeyedByTaskIdentifier is empty to release the session and self.
test


tewha:
A better way might be to have a session holder class between them. That way when the session holder was deallocated the session could be as well.

Kaspik:
Can be closed as this has been fixed I think.

PR #3961, opened 2017-05-29
fixed a bug when URLString like "https://www.google.com?"

ddhjy:


jshier:

PR #3962, opened 2017-05-30
Fix AFURLSessionManager security weakness

arttrick:
crashes when calling to a released block pointer

A security weakness was detected as a result of detection by Facebook's infer.
You can trigger a crash by sending nil to the url_session_manager_create_task_safely function as a debugger.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -39,13 +39,18 @@ static dispatch_queue_t url_session_manager_creation_queue() {
 }
 
 static void url_session_manager_create_task_safely(dispatch_block_t block) {
-    if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fixed_5871104061079552_bug) {
-        // Fix of bug
-        // Open Radar:http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
-        // Issue about:https://github.com/AFNetworking/AFNetworking/issues/2093
-        dispatch_sync(url_session_manager_creation_queue(), block);
+    
+    if(block != nil) {
+        if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fixed_5871104061079552_bug) {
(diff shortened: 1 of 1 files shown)


AFNetworking/AFURLSessionManager.m
```diff
@@ -39,13 +39,18 @@ static dispatch_queue_t url_session_manager_creation_queue() {
 }
 
 static void url_session_manager_create_task_safely(dispatch_block_t block) {
-    if (NSFoundationVersionNumber < NSFoundationVersionNumber_With_Fixed_5871104061079552_bug) {
-        // Fix of bug
-        // Open Radar:http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
-        // Issue about:https://github.com/AFNetworking/AFNetworking/issues/2093
-        dispatch_sync(url_session_manager_creation_queue(), block);
+    
+    if(block != nil) {
```
SlaunchaMan: Can you add a space after the `if` here?

AFNetworking/AFURLSessionManager.m
```diff
@@ -39,13 +39,18 @@ static dispatch_queue_t url_session_manager_creation_queue() {
...
+            // Fix of bug
+            // Open Radar:http://openradar.appspot.com/radar?id=5871104061079552 (status: Fixed in iOS8)
+            // Issue about:https://github.com/AFNetworking/AFNetworking/issues/2093
+            dispatch_sync(url_session_manager_creation_queue(), block);
+        } else {
+            block();
+        }
     } else {
-        block();
+        return;
```
SlaunchaMan: This entire `else` statement can be removed, since it’s just calling `return`.

SlaunchaMan:
This could use a test that verifies that calling the method with a `NULL` block doesn’t crash.

arttrick:
Good morning.
I'm sorry for late reply.
Can I add and reflect whitespace?
 
Have a nice day
-----Original Message-----

PR #3979, opened 2017-06-28
Add Alamofire’s security policy test suite (including certificates) to AFNetworking’s

SlaunchaMan:
The AFSecurityPolicyTests were using certificates that have since expired. The HTTPBin certificates transitioned from `*.httpbin.org` to `httpbin.org`, making it less useful for wildcard testing. To that end, I’ve copied in the certificates from the Alamofire project and converted the `ServerTrustPolicy` tests to new tests on `AFSecurityPolicy`.

This makes the test suites verify the same sets of APIs using the same certificates, where possible. I currently have a question about how the SSL pinning feature is *supposed* to work, so I’ve left some tests commented out. For now, this should get us a well-tested suite from AFNetworking with much longer-lived certificates.


jshier:
💯 👍 
Looks great but I'd want @cnoon to review it.

Tests/Tests/AFSecurityPolicyLegacyCertificates.m
```diff
@@ -0,0 +1,122 @@
...
+}
+
+SecTrustRef AFUTHTTPBinOrgServerTrust() {
+    NSString *bundlePath = [[NSBundle bundleForClass:NSClassFromString(@"AFSecurityPolicyTests")] resourcePath];
+    NSString *serverCertDirectoryPath = [bundlePath stringByAppendingPathComponent:@"HTTPBinOrgServerTrustChain"];
+    
+    return AFUTTrustChainForCertsInDirectory(serverCertDirectoryPath);
+}
+
+//static SecTrustRef AFUTGoogleComServerTrustPath1() {
```
jshier: Why are these commented out? If we aren't testing against the Google certs they should be deleted, unless you're going to bring them back?

jshier:
Too far out of date to merge.

PR #3983, opened 2017-07-03
Task created in a session that has been invalidate

vinhsteven:
The application will be crashed when it loses internet connection suddenly. We'll see the log in console something like this: 

`Attempted to create a task in a session that has been invalidated
2016-07-02 16:52:57.354 FlydeApp[19981:191244] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Task created in a session that has been invalidate`

My solution is define another variable: invalidateSession and this value will be referenced from current session whenever the session in AFURLSessionManager is invalidated. Then, before create a data task, I will compare the current session with the invalidateSession variables to make sure no data task is created in invalidated session. I tested this solution on my own project, and it works perfectly. Hope it will release a lot of headache for others as well.

### DIFF
--- a/AFNetworking/AFURLSessionManager.m
+++ b/AFNetworking/AFURLSessionManager.m
@@ -719,7 +719,14 @@ static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofi
 
     __block NSURLSessionDataTask *dataTask = nil;
     url_session_manager_create_task_safely(^{
-        dataTask = [self.session dataTaskWithRequest:request];
+//        dataTask = [self.session dataTaskWithRequest:request];
+        /** vinhsteven: fixed crash when lose connection suddenly.
+         the invalidatedSession was assigned when session is invalidated in - (void)URLSession:(NSURLSession *)session
+         didBecomeInvalidWithError:(NSError *)error
+         */
+        if (self.session != self.invalidatedSession) {
+            dataTask = [self.session dataTaskWithRequest:request];
+        }
     });
 
     [self addDelegateForDataTask:dataTask uploadProgress:uploadProgressBlock downloadProgress:downloadProgressBlock completionHandler:completionHandler];
@@ -914,6 +921,9 @@ static NSString * const AFNSURLSessionTaskDidSuspendNotification = @"com.alamofi
 - (void)URLSession:(NSURLSession *)session
 didBecomeInvalidWithError:(NSError *)error
 {
+    /** vinhsteven: fixed crash when lose connection suddenly. */
+    self.invalidatedSession = session;
+    
     if (self.sessionDidBecomeInvalid) {
         self.sessionDidBecomeInvalid(session, error);
     }
--- a/AFNetworking/AFURLSessionManager.h
+++ b/AFNetworking/AFURLSessionManager.h
@@ -95,6 +95,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly, nonatomic, strong) NSURLSession *session;
 
 /**
+ Invalidate session
+ */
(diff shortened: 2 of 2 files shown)


SlaunchaMan:
Thanks for the PR! A couple of comments:

- Can you merge in `master`? We just fixed the failing tests there.
- Can you add tests that show the breaking behavior of the existing code?
- What happens when the session becomes invalidated? Does the request just never happen? Should we instead create a new session when the current one is invalidated?

vinhsteven:
Hi @SlaunchaMan, 
1. I can't merge this pull request since I am not authorized to merge it.
2. What do you mean about adding tests here? In my case, I send multiple requests in background to retrieve data from server after logging in my application, then I interrupted the Internet connection while it hadn't finished reloading the main view. It crashes the app. I traced back and realized the crash come from this line: 

`url_session_manager_create_task_safely(^{ 
         dataTask = [self.session dataTaskWithRequest:request];
});`

So, my suspect was it tried to create a task in self.session, but this one was invalidated.

3. If the session become invalidate, the request doesn't do anything else. I think we shouldn't create any new session in this case, because most of the time this case only happens when you lose internet connection. What can you do when you lose your internet connection?

SlaunchaMan:
Hi @vinhsteven! I didn't mean to merge this PR into `master`, I meant to update this branch with the latest changes from `master` so tests pass again.

What would be ideal here is some tests that demonstrate the behavior of the class when the session is invalidated. You’d likely need to create a mock session that you can invalidate.

As for what we should do, iOS 11 introduces new APIs to automatically resume requests when the connection is reestablished. I don’t think that's mandatory for this PR, but it would be interesting to see if that is a viable option here.

jshier:
Losing connectivity shouldn't invalidate a session in the first place, so there's likely something else going on here, probably above AFNetworking itself. i.e. Something else happened that caused the session to be invalidated (an explicit call from outside, something automatic inside) but that didn't tear down the entire AFNetworking instance.

SlaunchaMan:
Sounds likely, @jshier. @vinhsteven, can you merge `master` back into this branch or rebase off of it?

xingheng:
Hi, team, please wait, I wonder this PR will get stuck in future requests. @SlaunchaMan @vinhsteven 

`nsurlsessiond` didn't accept the invalid request to throw a crash exception there, that's some incredible for me. What if the AFNetworking suppresses the invalidate request without any error message, it would be a potential bug in this library.

Highly suggest replacing the if statement with NSAssert macro there, it's more helpful.

AFNetworking/AFURLSessionManager.m
```diff
@@ -719,7 +719,14 @@ - (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request
 
     __block NSURLSessionDataTask *dataTask = nil;
     url_session_manager_create_task_safely(^{
-        dataTask = [self.session dataTaskWithRequest:request];
+//        dataTask = [self.session dataTaskWithRequest:request];
+        /** vinhsteven: fixed crash when lose connection suddenly.
+         the invalidatedSession was assigned when session is invalidated in - (void)URLSession:(NSURLSession *)session
+         didBecomeInvalidWithError:(NSError *)error
+         */
+        if (self.session != self.invalidatedSession) {
```
xingheng: Highly suggest replacing the if statement with NSAssert macro there, it's more helpful.

kobe1941:
meet again, how to fixed it?

jshier:
Thanks for the work but we won't be accepting this PR.

kobe1941:
@jshier  Hi，i want to know,
how to fixed this crash ?
the nearly release version if already fixed it or not ?

PR #4005, opened 2017-07-19
Add - Custom HTTPHeaderField remove method.

tezpark:
* Add - Removes the key for the HTTP headers set in the request serializer.

It is a confusing implementation to add an HTTPHeaderField to the AFHTTPResponseSerializer and remove the value by putting nil.
So I added a function that removes the HTTPHeaderField itself.


jshier:
Thanks for the PR! However, it won't be merged as the current API matches `NSMutableURLRequest` and so its semantics should be familiar to most users.

PR #4017, opened 2017-08-07
fix memory leak issue

zhazhenwang:
try to fix memory leak issue in AFURLSessionManager module


jshier:
There doesn't seem to be any actual memory leak fixed here. If you're still seeing an issue, please post a new PR with your analysis.

PR #4023, opened 2017-08-17
Fixed one issue which is:

superyl:
For AFSecurityPolicy, in method AFServerTrustIsValid, if the first validation attempt is failed but result is kSecTrustResultRecoverableTrustFailure, try to recover first and if recover is successful, send YES.


jshier:
This is invalid, as no actual recovery can take place, and this would lead to accepting invalid certificates.

PR #4034, opened 2017-08-30
Allow extension api only

BlueVirusX:
This PR fixes warnings when using AFNetworking in app extensions.


jshier:
Unfortunately this setting is incompatible when using a variety of AFNetworking extensions. To support it the library would need extensive refactoring to properly separate the different extensions.

PR #4041, opened 2017-09-06
AFNetworking: Fix static analyzer issues.

Ner1Co:


Ner1Co:
For some reason, 4a49ba0 solved the static analyzer alone.
Should we add an assert anyhow?

Ner1Co:
@dlahyani @StatusReport

PR #4051, opened 2017-09-18
Exposes C function prototype

ashfurrow:
Good morning! I came across a use case in unit testing (see [this discussion](https://github.com/artsy/eigen/pull/2423#issuecomment-330227309) for specifics) when accessing the `AFJSONObjectByRemovingKeysWithNullValues` would be really useful. I'd like to propose exposing this function publicly so that network response stubbing in unit tests can behave as closely as possible to network response serialization in production. Let me know if I can clarify anything – thanks!


ashfurrow:
🎉

PR #4052, opened 2017-09-18
Modifies AFJSONObjectByRemovingKeysWithNullValues to remove NSArray null objects

ashfurrow:
I added a test case that causes `AFJSONObjectByRemovingKeysWithNullValues` to return a JSON object containing `NSNull`s in arrays, and added check against `NSNull` before adding to the response array. Let me know if I can clarify anything. You can see a real-world test case here: https://github.com/artsy/eigen/pull/2423/files#diff-99f40649a0fc5dfc78a7014cefddddc5R104


ashfurrow:
Not sure about the CI failures – can anyone point me in the right direction?

SlaunchaMan:
Seems reasonable. The CI failures were probably from outdated test certificates; testing locally.

PR #4054, opened 2017-09-21
fix the bug of isReachable method is ineffective

x5forever:
In the absence of calling stopMonitoring method, the methods of isReachableXXX is ineffective.


x5forever:
In the absence of calling stopMonitoring method, the isReachable method is ineffective

jshier:
Thanks for the PR! However, due to its age it will not be merged. If the issue is still occurring feel free to submit it again.

PR #4056, opened 2017-09-22
fix the bug of isReachable method is ineffective, by x5.

x5forever:


jshier:
Thanks for the PR! Due to its age I'm closing it. If the issue is still occurring feel free to submit a new PR.

PR #4065, opened 2017-09-26
[Thread Race] removeObjectForKey: in NSDictionary.

RbBtSn0w:
when self.mergedTasks[URLIdentifier], is happen in
dispatch_async(self.responseQueue) [DISPATCH_QUEUE_CONCURRENT], then
have other thread add object to mergedTasks.

In this case, remove responseQueue(DISPATCH_QUEUE_CONCURRENT), current
performance consumption can be ignored on response thread.


streeter:
I believe this was fixed in https://github.com/AFNetworking/AFNetworking/pull/4246, which means this PR could be closed.

jshier:
@streeter is correct. Thanks for the PR, but this has been fixed.

PR #4081, opened 2017-10-12
Fixed CLANG_ENABLE_CODE_COVERAGE flag so release can be made

MapaX:
Submissions to iTunes Connect (built with Xcode 9) are failing in processing with the following message:

Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

Part of the problem is fixed in Carthage, but the Code coverage needs to be set NO in project. Atleast now.
https://github.com/Carthage/Carthage/issues/2056

This Pull request will update the flags so you can build and release your app.

https://developer.apple.com/library/content/qa/qa1964/_index.html


SlaunchaMan:
Looks like this PR also has some unrelated changes that are causing a conflict. @MapaX, can you merge in `master`?

Kaspik:
Can be merged for release 3.3.0?

PR #4102, opened 2017-11-21
fix #1825 bug. Because you call ``` [super encodeWithCoder:coder]; ``…

yupliang:
fix #1825 bug. Because you call ``` [super encodeWithCoder:coder]; ``` in AFURLSessionManager.m, So the problem that NSURLSessionConfiguration does not confirms to NSCoding protocol still be there. And I think call super does nothing wrong, so I rmove the encoding code for configuration to parent class. Happily this also fix the warning for ` *** NSKeyedArchiver warning: replacing existing value for key 'sessionConfiguration'; probable duplication of encoding keys in class hierarchy `


SlaunchaMan:
Can you merge with the latest `master` to fix tests?

yupliang:
I got lost. why 2 failing checks.

yupliang:
Although apple delete NSCoding protocol from NSURLSessionConfiguration. The concrete sub class does conform to NSCoding. Unit test can't reach here. So I can't let the unit test code reach the else branch. So I think it's better deleting the else branch.

SlaunchaMan:
I don't think deleting the `else` branch is the right answer there—I meant merging with the`master` branch to get some other fixes we’ve merged in.

yupliang:
Ok, now in my opinion, the code here only has one question that is subclass and super class use the same key to archive configuration which leads to a warning ```*** NSKeyedArchiver warning: replacing existing value for key 'sessionConfiguration'; probable duplication of encoding keys in class hierarchy ```. And preserve the else branch here doesn't leading to any problem. And the warning is not a real problem here, So if you want to ignores it, just feel free to close the pr.

jshier:
Tests around this pass, so I think we're good here.

PR #4113, opened 2017-12-15
Support for setting headers per request when using AFHTTPSessionManager.

stnslw:
It would be nice to have an option to set HTTP headers per request when using AFHTTPSessionManager. 
Currently it's only possible to set the default headers on the request serializer which are used for every request done by the http session manager instance. The problem occurs when you would like to send different headers for different requests which cannot be done in a clean way using AFHTTPSessionManager.
Setting headers per request can only be implemented by using AFURLSessionManager or by subclassing/extending AFHTTPSessionManager, which seems to be inconvenient. 
Also Alamofire library for Swift supports the `headers per request`. It uses a bit different API (without the manager), however, it still seems to be inconsistent.

What do you think?


tpot:
Would love to see this patch merged.  I'm in the process of reinventing this feature right now.

SlaunchaMan:
I’m going to look at this in more depth next for inclusion into a future 3.3.0 release—but on first inspection it looks good!

PR #4131, opened 2018-01-10
UIImageView+AFNetworking: Prevent stuck state for malformed urlRequest

aduflo:
Addresses issue within `setImageWithURLRequest:placeholderImage:success:failure:`.

Issue:
If a `NSURLRequest` object is initialized with a malformed url (eg. passing an uninitialized string into `URLWithString:`), and is passed into `setImageWithUrlRequest:`, neither blocks (`success` or `failure`) get fired. When the method encounters a `nil` `URL` property of the `NSURLRequest` argument, `cancelImageDownloadTask` is called, but it's functionality depends on the existence of `af_activeImageDownloadReceipt`. In this case `af_activeImageDownloadReceipt` is `nil`, as the early return prevents a download task from starting, thus leaving the method caller waiting.

Proposal:
Replace `cancelImageDownloadTask` with the firing of the `failure` block.


aduflo:
@SlaunchaMan from what I can tell, build job #5 failed due to a `fastlane` error: https://travis-ci.org/AFNetworking/AFNetworking/jobs/328142514

I'd restart the build job but do not have the proper permissions to do so; would you be able to restart the build job?

SlaunchaMan:
Restarted!

aduflo:
@SlaunchaMan got my fork UTD w/ master, and pushed a merge into the PR and saw all build jobs failing

it seems master has a handful of failing tests:
screen shot 2018-02-15 at 10 20 17 am

SlaunchaMan:
Looks like some certificates expired that we use for tests. I'll update those shortly!

SlaunchaMan:
Tests have been updated. Once you merge in `master` and the tests pass this is good to go!

aduflo:
`master` merged and seein' ✅s. 👍

PR #4133, opened 2018-01-12
Bug fix

shang1219178163:
PHP back to me  AESEncrypt's Object,and you know!


SlaunchaMan:
This is inside of `AFJSONResponseSerializer`’s serialization methods, which expect a JSON object. If you need to parse a string, consider using a different response serializer.

PR #4146, opened 2018-02-06
make it better!

shang1219178163:
PHP back to me  encrypted strings,and you know!


Coeur:
This is wrong. If you don't receive a JSON, then don't use AFJSONResponseSerializer.

Issue #4027, opened 2017-08-22
AFNetworkReachabilityManager.h generates warning "Conflicting nullability specifier on return types"


pouriaalmassi:
AFNetworkReachabilityManager.h generates a warning stemming from conflicting definitions of `init`. Specifically the following definitions conflict. 

AFNetworkReachabilityManager.h's `- (nullable instancetype)init NS_UNAVAILABLE;` conflicts with NSObject.h's `- (instancetype)init`.

Warning:  `AFNetworkReachabilityManager.h:115:1: Conflicting nullability specifier on return types, 'nullable' conflicts with existing specifier 'nonnull'`.

Environment
- Xcode Version 8.3.3 (8E3004b)
- iOS 10.3
- AFNetworking Version: 3.1.0
- macOS Version 10.12.6 (16G29)
- Installed via Cocoapods.

Jaron-Lowe:
I am also experiencing this issue in one of my projects.

**Environment**
- Xcode Version: 9.1 (9B55)
- iOS Version: 11.1
- Deployment Target: 10.0
- AFNetworking Version: 3.1.0
- macOS Version: 10.13.1
- Installed: Manually

sspitzer:
I'm seeing this too AFNetworking Version: 3.2.0, this silences the warning:

```
- (nullable instancetype)init NS_UNAVAILABLE;
```

to:

```
- (instancetype)init NS_UNAVAILABLE;
```

racer1988:
This is still happening :)

But I guess AFNetworking is not maintained anymore?

PR #4154, opened 2018-02-13
Remove conflicting nullable specifier on init

nrbrook:
The nullable specifier on init conflicts with the nonnull specifier on NSObject which generates a warning. This change removes it and throws an exception if init is used.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.h
+++ b/AFNetworking/AFNetworkReachabilityManager.h
@@ -108,11 +108,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithReachability:(SCNetworkReachabilityRef)reachability NS_DESIGNATED_INITIALIZER;
 
 /**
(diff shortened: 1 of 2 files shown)


jklausa:
gentlest of pings @SlaunchaMan — would you mind looking at it? As it currently stands, when using `AFNetworking` as a CocoaPod, it generates a bunch of warnings in client projects:
https://github.com/wordpress-mobile/WordPress-iOS/pull/8660

Looks like there was #4027 filed about it.

SlaunchaMan:
Hi @jklausa! Can you merge `master` in to get the tests passing?

jklausa:
@SlaunchaMan I'm not the original author of this patch, but I took the liberty of cherry-picking it onto master and filing another one — #4163, in order to save time :) 

Hopefully that's ok with you @nrbrook — let me know if I should close my PR if you'd prefer that.

nrbrook:
Either is fine, shall I close this then?

PR #4182, opened 2018-03-16
Remove conflicting nullable specifier on init

SlaunchaMan:
See #4154—this is just cherry-picking it onto `master`. Closes #4163. Closes #4154.


Coeur:
Thanks for looking at it, @SlaunchaMan, albeit I can't see what's the issue about simply merging either #4163 or #4154.

SlaunchaMan:
Yep, just trying to get the build bot happy.

PR #4190, opened 2018-03-28
Check Security Policy only if baseURL is set

mojtabacazi:
There is no reason to assert on security policy if there is no baseURL. Doing that will break existing code that are indeed using https, but baseURL is not being used

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.m
+++ b/AFNetworking/AFHTTPSessionManager.m
@@ -101,7 +101,7 @@
 @dynamic securityPolicy;
 
 - (void)setSecurityPolicy:(AFSecurityPolicy *)securityPolicy {
-    if (securityPolicy.SSLPinningMode != AFSSLPinningModeNone && ![self.baseURL.scheme isEqualToString:@"https"]) {
+    if (securityPolicy.SSLPinningMode != AFSSLPinningModeNone && self.baseURL && ![self.baseURL.scheme isEqualToString:@"https"]) {
         NSString *pinningMode = @"Unknown Pinning Mode";
         switch (securityPolicy.SSLPinningMode) {
             case AFSSLPinningModeNone:        pinningMode = @"AFSSLPinningModeNone"; break;
--- a/Tests/Tests/AFHTTPSessionManagerTests.m
+++ b/Tests/Tests/AFHTTPSessionManagerTests.m
@@ -550,6 +550,18 @@
     XCTAssertThrowsSpecificNamed(manager.securityPolicy = securityPolicy, NSException, @"Invalid Security Policy");
 }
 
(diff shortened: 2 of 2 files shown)


SlaunchaMan:
Instead of removing tests that break when you make this change, can you change the tests to instead pass with the new code?

mojtabacazi:
@SlaunchaMan That's a good idea. Done.

mojtabacazi:
@SlaunchaMan Do want me add more tests? and documentation?

SlaunchaMan:
@mojtabacazi I need to think about this PR some more, as it has pretty significant security implications. Can you explain a situation where you would need to specify a non-default security policy but have no `baseURL`? The only thing I can think of is wildcard domains in pinned certificates, and I'm not sure how common that is.

mojtabacazi:
@SlaunchaMan There probably is a baseURL but it's not necessary set in SessionManager since it was always optional, in our case we have our own URL constructor that has baseURL, etc.

SlaunchaMan:
@mojtabacazi But are you using a security policy? The first part of the conditional is `securityPolicy.SSLPinningMode != AFSSLPinningModeNone`, which means this code won’t execute if you don’t specify a separate security policy, as the default policy is `AFSSLPinningModeNone`.

mojtabacazi:
@SlaunchaMan We use `AFSSLPinningModePublicKey`

SlaunchaMan:
@mojtabacazi Is your public key using a wildcard? I would normally expect that the `baseURL` would be set to match the domain name used in the certificate.

mojtabacazi:
@SlaunchaMan It matches the baseURL we use to construct NSURLs.

SlaunchaMan:
This PR needs the following tests:

- A request with a `nil` `baseURL`, a pinning mode that’s not `AFSSLPinningModeNone`, and a URL that begins with "https" should succeed iff the security policy succeeds.
- A request with a `nil` `baseURL`, a pinning mode that’s not `AFSSLPinningModeNone`, and a URL that does not begin with "https" should fail with a descriptive error.

Additionally, the documentation for `AFHTTPSessionManager` should be updated to indicate that if you use `nil` for your `baseURL`, then your requests must be secure URLs that match the certificate or public key in the security policy.

jshier:
If you can add the suggested tests, this should be mergeable.

mojtabacazi:
> If you can add the suggested tests, this should be mergeable.

@jagill Github somehow lost the link to the repo I submitted this PR from, not sure how to link it back. If you know how, otherwise I'm going to create a new PR with requested changes.

jshier:
@mojtabacazi No idea about that, feel free to open a new PR.

PR #4201, opened 2018-04-13
Update code coverage settings

SlaunchaMan:
This PR tries to improve the quality and reduce the noise of the code coverage integration by:

- Ignoring the `Tests` folder since we don’t care about coverage of test code,
- Running the code coverage on multiple OS versions to capture availability branching, and
- Running the code coverage on multiple OSes to capture OS branching.


frouo:
Hi, I am adding codecov to my pod and was looking around how it's done in other pods.
I just noticed that according to codecov it's not recommended to prefix the config file name with a dot ([reference](https://docs.codecov.io/docs/codecov-yaml#can-i-name-the-file-codecovyml)).

> Can I name the file .codecov.yml?
> No. Starting the file with a period can cause issues detecting and reading it.

I don't know how it might affect `AFNetworking` codecov in the future. Just warning in case it can help 🙂

PR #4231, opened 2018-06-14
Code fix download image

jklin559:
when the request method is "post",it will return the same key for different requests


SlaunchaMan:
Can you add tests that demonstrate the problem you’re fixing here?

jshier:
Thanks for the PR! However, while there may be a caching issue here, appending the fully body data as the key value is not the correct solution. Feel free to submit another PR with a better solution though!

Issue #3984, opened 2017-07-04
Race condition in -[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]


aleksgapp:
v3.1

```
==================
WARNING: ThreadSanitizer: race on a library object (pid=20259)
  Read-only access of object NSMutableDictionary at 0x7b0c0010d0d0 by thread T19:
    #0 -[__NSDictionaryM objectForKeyedSubscript:]
  * #1 __78-[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]_block_invoke_2.190

  Previous mutating access of object NSMutableDictionary at 0x7b0c0010d0d0 by thread T9:
    #0 -[__NSDictionaryM removeObjectForKey:]
  * #1 -[AFImageDownloader removeMergedTaskWithURLIdentifier:]
    #2 __61-[AFImageDownloader safelyRemoveMergedTaskWithURLIdentifier:]_block_invoke
    #3 __tsan::invoke_and_release_block(void*)
    #4 _dispatch_client_callout
    #5 -[AFImageDownloader safelyRemoveMergedTaskWithURLIdentifier:]
    #6 __78-[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]_block_invoke_2.190 

  Issue is caused by frames marked with "*".

  Thread T19 (tid=6605263, running) is a GCD worker thread
  Thread T9 (tid=6605223, running) is a GCD worker thread

SUMMARY: ThreadSanitizer: race on a library object <redacted> in __78-[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]_block_invoke_2.190
==================
```

```objc
T19: AFImageDownloaderMergedTask *mergedTask = self.mergedTasks[URLIdentifier];     <--- AFImageDownloader.m:241
---
T9: [self.mergedTasks removeObjectForKey:URLIdentifier];   <--- AFImageDownloader.m:336
```

RbBtSn0w:
[https://github.com/AFNetworking/AFNetworking/pull/4065](Pull)

Have a choose for you.

billburgess:
+1 on this issue.

malinois:
+1 on this issue

streeter:
I believe this was fixed by https://github.com/AFNetworking/AFNetworking/pull/4246, which means this issue could be closed.

Issue #3876, opened 2017-02-15
Crash in [AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]


stephenyao:
Currently using AFNetworking 3.1.0, we are getting a crash in our app when an image download task is happening. Below is the stack trace:

```
Crashed: com.alamofire.imagedownloader.responsequeue-2FA8BE81-6D24-4219-B9A1-80ECB8E1F0CF
0  libobjc.A.dylib                0x1810c6f70 objc_msgSend + 16
1  CoreFoundation                 0x182564148 CFStringCompareWithOptionsAndLocale + 480
2  Foundation                     0x1830a3ac4 -[NSString isEqualToString:] + 76
3  Foundation                     0x1830a3a5c -[NSString isEqual:] + 68
4  CoreFoundation                 0x1825538dc CFEqual + 456
5  CoreFoundation                 0x182556a44 -[__NSCFString isEqual:] + 312
6  CoreFoundation                 0x18255f988 -[__NSDictionaryM objectForKey:] + 148
7  realestate.com.au              0x10054bd4c __78-[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:]_block_invoke_2.190 + 4301405516
8  libdispatch.dylib              0x1815061fc _dispatch_call_block_and_release + 24
9  libdispatch.dylib              0x1815061bc _dispatch_client_callout + 16
10 libdispatch.dylib              0x181512f94 _dispatch_continuation_pop + 576
11 libdispatch.dylib              0x181512c58 _dispatch_async_redirect_invoke + 628
12 libdispatch.dylib              0x18151634c _dispatch_root_queue_drain + 572
13 libdispatch.dylib              0x1815160ac _dispatch_worker_thread3 + 124
14 libsystem_pthread.dylib        0x18170f2a0 _pthread_wqthread + 1288
15 libsystem_pthread.dylib        0x18170ed8c start_wqthread + 4
```

The final message from console log was:

`#0
Crashed: com.alamofire.imagedownloader.responsequeue-2FA8BE81-6D24-4219-B9A1-80ECB8E1F0CF
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000c62f0beb8`

Please help

PR #4246, opened 2018-07-12
Safely get mergedTasks On responseQueue

bbeversdorf:
There is a race condition in -[AFImageDownloader downloadImageForURLRequest:withReceiptID:success:failure:] that can cause a crash.
#3876

The race condition can be seen with the Xcode Thread Sanitizer on and the project downloading and cancelling images quickly. 
In particular I noticed this downloading images for UITableViewCells and scrolling through the tableView.
#3984 

I think the main issue here is that `mergedTasks` is always updated on `synchronizationQueue` and it is being accessed on `responseQueue`. So there could be a change to the dictionary on a different thread.

### DIFF
--- a/UIKit+AFNetworking/AFImageDownloader.m
+++ b/UIKit+AFNetworking/AFImageDownloader.m
@@ -254,7 +254,7 @@
                        completionHandler:^(NSURLResponse * _Nonnull response, id  _Nullable responseObject, NSError * _Nullable error) {
                            dispatch_async(self.responseQueue, ^{
                                __strong __typeof__(weakSelf) strongSelf = weakSelf;
-                               AFImageDownloaderMergedTask *mergedTask = strongSelf.mergedTasks[URLIdentifier];
+                               AFImageDownloaderMergedTask *mergedTask = [strongSelf safelyGetMergedTask: URLIdentifier];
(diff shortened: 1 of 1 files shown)


UIKit+AFNetworking/AFImageDownloader.m
```diff
@@ -254,7 +254,7 @@ - (nullable AFImageDownloadReceipt *)downloadImageForURLRequest:(NSURLRequest *)
                        completionHandler:^(NSURLResponse * _Nonnull response, id  _Nullable responseObject, NSError * _Nullable error) {
                            dispatch_async(self.responseQueue, ^{
                                __strong __typeof__(weakSelf) strongSelf = weakSelf;
-                               AFImageDownloaderMergedTask *mergedTask = strongSelf.mergedTasks[URLIdentifier];
+                               AFImageDownloaderMergedTask *mergedTask = [strongSelf safelyGetMergedTask: URLIdentifier];
```
SlaunchaMan: Can you remove the space after the `:` here?

bbeversdorf: Certainly! I will get that up when I can.

Jackylee1992:
👍🏻

PR #4248, opened 2018-07-13
Add an ability that notification-observer know which reachability it belongs

LMsgSendNilSelf:
there exists one problem [AFNetworkReachabilityManager shareManager] is singleton，but it also support that create other AFNetworkReachabilityManager by method ‘managerForDomain’ or ’managerForAddress‘. the two case is opposite.
so if network change, trigger posting AFNetworkingReachabilityDidChangeNotification , receiver can not know which is the actully sender，it will get multiple notification，then the receiver notification method would call multiple times, maybe cause continuously unexpectedly result.

### DIFF
--- a/AFNetworking/AFNetworkReachabilityManager.h
+++ b/AFNetworking/AFNetworkReachabilityManager.h
@@ -151,6 +151,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setReachabilityStatusChangeBlock:(nullable void (^)(AFNetworkReachabilityStatus status))block;
 
+/**
+ Returns the reachability of current networkReachabilityManager.
+ */
+- (SCNetworkReachabilityRef)currentNetworkReachability;
+
 @end
 
 ///----------------
--- a/AFNetworking/AFNetworkReachabilityManager.h
+++ b/AFNetworking/AFNetworkReachabilityManager.h
@@ -151,11 +151,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setReachabilityStatusChangeBlock:(nullable void (^)(AFNetworkReachabilityStatus status))block;
 
-/**
- Returns the reachability of current networkReachabilityManager.
- */
-- (SCNetworkReachabilityRef)currentNetworkReachability;
-
 @end
 
 ///----------------
--- a/AFNetworking/AFNetworkReachabilityManager.m
+++ b/AFNetworking/AFNetworkReachabilityManager.m
@@ -78,7 +78,7 @@ static AFNetworkReachabilityStatus AFNetworkReachabilityStatusForFlags(SCNetwork
  * a queued notification (for an earlier status condition) is processed after
  * the later update, resulting in the listener being left in the wrong state.
  */
-static void AFPostReachabilityStatusChange(SCNetworkReachabilityFlags flags, AFNetworkReachabilityStatusBlock block) {
+static void AFPostReachabilityStatusChange(SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags, AFNetworkReachabilityStatusBlock block) {
     AFNetworkReachabilityStatus status = AFNetworkReachabilityStatusForFlags(flags);
     dispatch_async(dispatch_get_main_queue(), ^{
         if (block) {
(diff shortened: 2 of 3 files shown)


AFNetworking/AFNetworkReachabilityManager.m
```diff
@@ -78,20 +78,20 @@ static AFNetworkReachabilityStatus AFNetworkReachabilityStatusForFlags(SCNetwork
  * a queued notification (for an earlier status condition) is processed after
  * the later update, resulting in the listener being left in the wrong state.
  */
-static void AFPostReachabilityStatusChange(SCNetworkReachabilityFlags flags, AFNetworkReachabilityStatusBlock block) {
+static void AFPostReachabilityStatusChange(SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags, AFNetworkReachabilityStatusBlock block) {
```
SlaunchaMan: Is there an advantage to using the `SCNetworkReachabilityRef` directly here, rather than the `AFNetworkReachabilityManager`? That way we can keep the abstraction around the SCNetwork stuff intact, but still tell the user which manager is posting the notification.

AFNetworking/AFNetworkReachabilityManager.h
```diff
@@ -151,6 +151,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setReachabilityStatusChangeBlock:(nullable void (^)(AFNetworkReachabilityStatus status))block;
 
+/**
+ Returns the reachability of current networkReachabilityManager.
+ */
+- (SCNetworkReachabilityRef)currentNetworkReachability;
```
SlaunchaMan: It looks like you’re not actually using this anywhere in the code; how are you anticipating this to be used?

AFNetworking/AFNetworkReachabilityManager.m
LMsgSendNilSelf: It's that we can‘t obtain AFNetworkReachabilityManager directly to  post.
 In AFNetworkReachabilityCallback function,  correlated object for AFNetworkReachabilityManager  is just SCNetworkReachabilityRef, .
AFNetworkReachabilityManager and SCNetworkReachabilityRef is one to one.

if  using AFNetworkReachabilityManager,
1、change AFNetworkReachabilityStatusBlock returntype from void to AFNetworkReachabilityManager
2、create the relationship map from SCNetworkReachabilityRef to  AFNetworkReachabilityManager，when obain current AFNetworkReachabilityManager, check the map
the two type,whick is better? or use SCNetworkReachabilityRef directly, thanks

AFNetworking/AFNetworkReachabilityManager.h
LMsgSendNilSelf: for example
...
- (void)method {
self.manager = [AFNetworkReachabilityManager managerForDomain:@"xxx"];
 [[NSNotificationCenter defaultCenter] addObserver:self
                                                 selector:@selector(networkReachabilityChanged:)
                                                     name:AFNetworkingReachabilityDidChangeNotification
                                                   object:nil];
}
...
- (void)networkReachabilityChanged:(NSNotification *)notify {
    AFNetworkReachabilityManager *selfManager = self.manager;
    id owneredReachabilityRef = notification.object;
    id currentReachabilityRef = (__bridge id)[selfManager currentNetworkReachability];
    
    if (![owneredReachabilityRef isEqual:currentReachabilityRef]) {
        return;
    }
   //your code
...
}

LMsgSendNilSelf: if we not use  SCNetworkReachabilityRef in  AFNetworkReachabilityManager.h,
Instead of '- (SCNetworkReachabilityRef)currentNetworkReachability;'
we also can override:
- (BOOL)isEqual:(id)object

AFNetworking/AFNetworkReachabilityManager.m
LMsgSendNilSelf: @SlaunchaMan

LMsgSendNilSelf:
@SlaunchaMan  ~ has update

Tests/Tests/AFNetworkReachabilityManagerTests.m
```diff
@@ -125,4 +125,26 @@ - (void)testDomainReachabilityBlock {
     [self verifyReachabilityStatusBlockGetsCalledWithManager:self.domainReachability];
 }
 
+- (void)testDifferentReachabilityManagerNotificationPostsAndGets {
```
SlaunchaMan: Instead of adding these tests, it should be sufficient to change the previous tests to include the correct parameter for `object:` in the `expectationForNotification:object:handler:` method call.

LMsgSendNilSelf:
@SlaunchaMan ~ has update,

PR #4253, opened 2018-07-21
Logical repair

ccworld1000:
Logical repair : self.mutableObservedChangedKeyPaths maybe is empty But the loop is still in execution
If there is no element, it should skip


jshier:
Thanks for the PR! However, this change doesn't do anything, as trying to loop over empty collections doesn't do anything anyway, so the check is unnecessary.

PR #4255, opened 2018-07-22
CCAutoTag : Chinese README translate

ccworld1000:
Chinese README translate


jshier:
Thanks for the PR! However, since we have no way to verify the contents of such a README, and we can't maintain it, we won't be merging it. Thanks anyway!

Issue #4140, opened 2018-01-30
Crash in CoreFoundation - CFDictionaryGetValue


bondarenko95:
We have a crash in the app. We using the AFNetworking 3.0. 

There are logs: 

> Crashed: com.apple.NSURLSession-work
> 0  CoreFoundation                 0x228f7cdc CFDictionaryGetValue + 27
> 1  Foundation                     0x231eeb5b _NSSetLongLongValueAndNotify + 54
> 2  Foundation                     0x231eeb5b _NSSetLongLongValueAndNotify + 54
> 3  CFNetwork                      0x22f05251 -[__NSCFLocalDataTask _onqueue_didReceiveResponse:completion:] + 68
> 4  CFNetwork                      0x22f48789 __61-[__NSCFURLLocalSessionConnection _didReceiveResponse:sniff:]_block_invoke + 96
> 5  CFNetwork                      0x22f473c3 -[__NSCFURLLocalSessionConnection withDelegateAndPendingCompletion:] + 190
> 6  CFNetwork                      0x22f4871b -[__NSCFURLLocalSessionConnection _didReceiveResponse:sniff:] + 374
> 7  CFNetwork                      0x22f4aa67 SessionConnectionLoadable::_loaderClientEvent_DidReceiveResponse(_CFURLResponse*) + 170
> 8  CFNetwork                      0x22fb1271 ___ZN19URLConnectionLoader26protocolDidReceiveResponseEP14_CFURLResponse_block_invoke + 24
> 9  libdispatch.dylib              0x2255d823 _dispatch_call_block_and_release + 10
> 10 libdispatch.dylib              0x2256a423 _dispatch_queue_drain$VARIANT$mp + 1758
> 11 libdispatch.dylib              0x22569a61 _dispatch_queue_invoke$VARIANT$mp + 284
> 12 libdispatch.dylib              0x2256c15d _dispatch_root_queue_drain + 396
> 13 libdispatch.dylib              0x2256bfcd _dispatch_worker_thread3 + 96
> 14 libsystem_pthread.dylib        0x22721b29 _pthread_wqthread + 1024
> 15 libsystem_pthread.dylib        0x22721718 start_wqthread + 8 

This issue occurs on the iPad 2 (iOS 9.3.5), iPad 3 (iOS 9.3.5), iPhone 6 Plus (iOS 9.3.0).

Can you please help resolve this issue? 
Thanks!

SlaunchaMan:
[...]

PR #4256, opened 2018-07-23
Fix crashes and memory leaks

Kaspik:
Good morning guys!

I have some improvements in AFNetowrking that I've done in November and was testing on our app since then. I would love to switch back to CocoaPods version from our fork so I'm opening PR for these Fabric crashes here.

- `invalidateSessionCancelingTasks` should also invalidate session as per name - this means the session has to be able to be allocated again not only in init method - `synchronized lazy loader` looks like best solution
- `session` in load can be taken from `sharedSession` - no reason to create new one by configuration
- `getTasksWithCompletionHandler` has to use `weak`/`strong`

Verified on app with almost 2 mil. active users daily since November 2017.


Kaspik:
Test `testRDAR17029580IsFixed` is failing and I'm not sure why.
It's test for iOS 7 but Travis CI is not even running iOS 7 tests anymore.
Are you going to drop iOS 7 support? Should we still keep the tests there with these fixes?

Kaspik:
Test testRDAR17029580IsFixed is failing and I'm not sure why. ❌
It's also green locally on my machine. ✅
It's test for iOS 7 but Travis CI is not even running iOS 7 tests anymore.
Are you going to drop iOS 7 support? Should we still keep the tests there with these fixes?

@kcharwood / @nikitahils - As you implemented this fix / tests 4 years ago (https://github.com/AFNetworking/AFNetworking/pull/3205) - any ideas why mine changes would affect this test?

SlaunchaMan:
Can you explain more about the crashes you see when calling `invalidateSessionCancelingTasks`? A session manager manages a session; once that session is invalidated, I would expect the session manager would also need to be recreated.

Kaspik:
@SlaunchaMan Specifically this change is memory leak fix, not crash fix.

The method `invalidateSessionCancelingTasks` is now also nilling the session as otherwise the session is not invalidated anytime at all and it's causing memory leak issue because session can then live even when it shouldn't.

We could probably leave the change only partially to change the session to lazy loader and nil it when it's called from other places in AFNetworking.
But in our case we tried and decided that the session can be invalidated anytime when there is nothing to do and all tasks are done and can be re-created when it's necessary to use it.

We are not seeing any issues since November 18th 2017 when we released our app with forked AFNetowrking and it also fixed a lot of crashes on `CFDictionaryGetValue` that has ben reported multiple times by multiple people here (like https://github.com/AFNetworking/AFNetworking/issues/4140).

As per your comment - session manager is managing the session. It can manage the session to be nil and then allocate it again, with similar configuration, but there is no need to create the manager itself again as the configuration of the manager would be the same and static.

SlaunchaMan:
Can you explain more about why you're calling `invalidateSessionCancelingTasks`? The way the API is described implies that there should be a 1:1 relationship between sessions and session managers; are you sure the leak is the session itself and not that you’re holding on to session managers for invalidated sessions?

Kaspik:
@SlaunchaMan I did some research in our app and performed some tests. It looks like the invalidate could be removed. Currently releasing to production to see if it's true or not - I'll get back to you next week about situation.

Definitely looking forward to merge this as we switched for 2 weeks back to stable last AFNetworking version on CocoaPods and our crash rate increased quite a lot.

Kaspik:
@SlaunchaMan Morning! Getting back to you as I promised. So current setup of this PR looks good and is stable on our app. (2mil+ daily users)

Let me know what do you think about merging this so we can move forward ideally also with CocoaPods release.

Kaspik:
@SlaunchaMan No worries.

Yes. The invalidate method is still called sometimes - especially if app is backgrounded for longer time or anything else - and the session can be then cleared out and created again with fresh config. Reverting back to not setting it to nil also bring back some issues on Fabric... 😒

SlaunchaMan:
@Kaspik That makes sense, but I’m wondering why you can’t simply recreate the session manager instead of recreating its session.

Kaspik:
Any update on merging this? If not, I'm going to switch back to our fork because crashes happening on current AFNetworking library are pretty bad (as you can see, it started when we switched back to no-fork version) - http://crashes.to/s/6daea4ae313

SlaunchaMan:
@Kaspik I think it’ll be OK to make this change. We should try to restrict this PR to making that one change—that is, allowing a session manager to recreate its session. I’ll make some more PR comments to that effect to remove some of the other stuff. Can you modify the `invalidateSessionCancelingTasks` method to take another `BOOL` parameter, something like `recreateSession`, and have that control whether or not the session is recreated? The implementation of `invalidateSessionCancelingTasks` would then call this new method with that parameter set to `NO`, so the behavior won’t change for existing code.

AFNetworking/AFURLSessionManager.m
```diff
@@ -404,8 +404,7 @@ + (void)load {
             7) If the current class implementation of `resume` is not equal to the super class implementation of `resume` AND the current implementation of `resume` is not equal to the original implementation of `af_resume`, THEN swizzle the methods
             8) Set the current class to the super class, and repeat steps 3-8
          */
-        NSURLSessionConfiguration *configuration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
-        NSURLSession * session = [NSURLSession sessionWithConfiguration:configuration];
+        NSURLSession * session = [NSURLSession sharedSession];
```
SlaunchaMan: Using the ephemeral session configuration here prevents some side effects that might happen on the shared session. Let’s leave this as it is. As we move to new Xcode versions, I suspect this code may be unnecessary, so it’ll probably be removed anyway.

AFNetworking/AFURLSessionManager.m
```diff
@@ -425,7 +424,6 @@ + (void)load {
         }
         
         [localDataTask cancel];
-        [session finishTasksAndInvalidate];
```
SlaunchaMan: As we revert the above code, this should be reverted too.

AFNetworking/AFURLSessionManager.m
```diff
@@ -556,6 +555,19 @@ - (void)dealloc {
 
 #pragma mark -
 
+- (NSURLSession *)session {
```
SlaunchaMan: Instead of making a new getter that automatically creates the session, let’s make a new `intiializeSession` method that creates the session with the saved session configuration and delegate queue, and call it from the initializer.

AFNetworking/AFURLSessionManager.m
```diff
@@ -717,6 +729,7 @@ - (void)invalidateSessionCancelingTasks:(BOOL)cancelPendingTasks {
     } else {
         [self.session finishTasksAndInvalidate];
     }
+    self.session = nil;
```
SlaunchaMan: Instead of setting the session to `nil`, to preserve existing behavior, let’s only recreate the session if the user explicitly passes `YES` to a new method that takes an additional parameter.

Tests/Tests/AFURLSessionManagerTests.m
```diff
@@ -158,19 +158,11 @@ - (void)testSessionTaskDoesReportMetrics {
     [self waitForExpectationsWithCommonTimeout];
 }
 
-// iOS 7 has a bug that may return nil for a session. To simulate that, nil out the
-// session and it will return nil itself.
-- (void)testFileUploadTaskReturnsNilWithBug {
+- (void)testSessionRecreatesAgain {
```
SlaunchaMan: When the other changes happen, let’s test the existing invalidation behavior as well as new invalidation behavior.

AFNetworking/AFURLSessionManager.m
```diff
@@ -533,17 +529,20 @@ - (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)config
     self.lock = [[NSLock alloc] init];
     self.lock.name = AFURLSessionManagerLockName;
 
+    __weak typeof(self) weakSelf = self;
```
SlaunchaMan: This is likely unnecessary; the `getTasksWithCompletionHandler` method is short-lived, and we probably want to keep the session manager around until it ends.

Kaspik: Because we are then assigning it to `__strong`, it explicitly holds the manager around until it ends (if we would use only `__weak`, then you are right and we would lose the manager. Without this change - self is deallocated during the enumeration and memory leaks appear.

AFNetworking/AFURLSessionManager.m
Kaspik: hmmm, let's think about this one. I can change the setter in `invalidateSessionCancelingTasks` to nil session only if user wants, but I would love to leave the lazy loader here for two reasons:
1) Lazy loader makes sure that if we nil the session, it get's recreated again and the behavior doesn't change
2) it fixes bug on iOS that you mentioned below - let's take a look on the test: `// iOS 7 has a bug that may return nil for a session. To simulate that, nil out the session and it will return nil itself` - this is wrong approach I think. If iOS 7 has bug where it can return nil, we should fix this bug by having lazy loader and not have the session nil at all - ever. So this is like 2 in 1 fix of `initialize` method + `iOS 7 bug fix`. What do you think?

AFNetworking/AFURLSessionManager.m
Kaspik: Ohh yes, let's just move out of iOS 7 and clean the whole code. 😄

Kaspik:
screenshot 2018-09-11 at 13 27 59
Okay, I have no idea what's going on there on Travis, but tests succeeded locally and I'm not gonna be fixing something that is not broken at all. 😄

AFNetworking/AFURLSessionManager.m
SlaunchaMan: Yeah, actually that would work. Since the default, current behavior wouldn’t `nil` out the session, the lazy loader wouldn’t affect existing users. Then we could clean up all of that documentation about the iOS 7 bug!

AFNetworking/AFURLSessionManager.m
SlaunchaMan: Right, I mean we could just use `self` inside the completion handler without fear. The completion handler is only going to maintain a strong reference to `self` while it’s executing; once it’s done, then that reference goes away. So we don’t need to do the `__weak`/`__strong` dance because this isn’t a retain cycle.

Kaspik: mmm, I still think there is problem with keeping self inside the closure and there is retain cycle. If it doesn't hurt, I would love to leave it there. :)

Kaspik:
@SlaunchaMan Okay, what are we gonna do about the tests to finish this up? I think we handled all comments.

PR #4280, opened 2018-09-03
fix a crash!  …

shang1219178163:
fix a crash!  …
if  ([parameters is [NSString class]]){

//code   crash!!!


shang1219178163:
fix a crash!

PR #4339, opened 2018-12-24
Changed the protocol method code order

HuginChen:
Make changes according to the sequence of system protocol methods


jshier:
Thanks for the PR, but we won't be merging merely stylistic changes. Thanks anyway!

PR #4349, opened 2019-01-11
Availibility and warnings.

hhanesand:
Add -Wextra, and fix resulting warnings.
Bump availability warnings to aggressive, fix warnings.


SlaunchaMan:
Can you merge in `master` so tests can succeed?

hhanesand:
@SlaunchaMan bump. Merged but erroring. Anything I can do about it?

jshier:
Thanks for the PR! More warnings are good, as are the availability annotations. However, adding the `__unused` annotations is rather unsightly and not desired. Please just turn that warning off. Otherwise I think this is a good addition.

PR #4360, opened 2019-02-22
Updated progress argument

mlostekk:
Updated the progress argument in the multipart example


jshier:
Thanks for the PR, but inline blocks are the preferred style for example code.

PR #4365, opened 2019-03-01
Fix podspec #trivial #important

Kaspik:
- Fix podspec as library version supports iOS 8+ (in project) but podspec says iOS 7+

Note: We should release new version with this fix I think.


Kaspik:
@SlaunchaMan I hope I'm not wrong with this issue but if I'm not it could potentially cause some issues?

SlaunchaMan:
Yep, I had planned this for 3.3.0 anyway—we don’t support below iOS 8 and neither do modern versions of Xcode.

PR #4393, opened 2019-05-07
Post a notification when finishing moving the downloaded file successfully.

xingheng:


jshier:
This PR is acceptable, but a test needs to be added to verify the behavior. If you can add them in the next few days these then it may be able to make the 4.0 release.

xingheng:
It seems the travis build failure is irrelevant with my changes, shall we skip that check result? BTW, please review my test case in the new commit, it passed for me.
@jshier

PR #4406, opened 2019-06-14
Fix `NetworkActivityIndicator` kvo call back will trigger twice bug.

kinarobin:
This is a bug when setter method manual KVO，this will trigger twice KVO call-back! 

1).  From Apple Doc:
```
Given a key that identifies a property (attribute, to-one relationship, or ordered or unordered to-many relationship), send -observeValueForKeyPath:ofObject:change:context: notification messages of kind NSKeyValueChangeSetting to each observer registered for the key, including those that are registered with other objects using key paths that locate the keyed value in this object. Invocations of these methods must always be paired.

The change dictionaries in notifications resulting from use of these methods contain optional entries if requested at observer registration time:
    - The NSKeyValueChangeOldKey entry, if present, contains the value returned by -valueForKey: at the instant that -willChangeValueForKey: is invoked (or an NSNull if -valueForKey: returns nil).
    - The NSKeyValueChangeNewKey entry, if present, contains the value returned by -valueForKey: at the instant that -didChangeValueForKey: is invoked (or an NSNull if -valueForKey: returns nil).
- (void)willChangeValueForKey:(NSString *)key;
- (void)didChangeValueForKey:(NSString *)key;
```

Since our code not use `-observeValueForKeyPath:ofObject:change:context:`, it isn't necessary to trigger KVO call-back manual.

2).  I noticed that [Alamofire/AlamofireNetworkActivityIndicator](https://github.com/Alamofire/AlamofireNetworkActivityIndicator/blob/master/Source/NetworkActivityIndicatorManager.swift) didn't use KVO manual, I think we should keep up with it.

3). Since there are not  expose `activityCount` property, It would be pointless to change `activityCount` value via KVC.


kinarobin:
@jshier   Done!

PR #4407, opened 2019-06-15
Optimizing cancel image download and let task to be canceled really.

kinarobin:
Now, I'm  using `AFImageDownloader` to download images, When I cancel task with image download receipt, I found the image is still downloaded and the imageCache store the image  successful. So once the task had been canceled, the network shouldn't download this image anymore. 

If each image is not cancelled correctly, it will cause each downloaded image to be decompressed, resulting in very tight memory.


kinarob:
The test case like this：

```Ruby
- (void)testThatCancelImageDownloadItShouldCancelImmediately {
    [self.downloader.imageCache removeAllImages];
    NSArray *imageURLStrings = @[
                                 @"https://secure.gravatar.com/avatar/5a105e8b9d40e1329780d62ea2265d8a?d=identicon",
                                 @"https://secure.gravatar.com/avatar/6a105e8b9d40e1329780d62ea2265d8a?d=identicon",
                                 @"https://secure.gravatar.com/avatar/7a105e8b9d40e1329780d62ea2265d8a?d=identicon",
                                 @"https://secure.gravatar.com/avatar/8a105e8b9d40e1329780d62ea2265d8a?d=identicon",
                                 @"https://secure.gravatar.com/avatar/9a105e8b9d40e1329780d62ea2265d8a?d=identicon"
                                 ];

    for (NSString *imageURLString in imageURLStrings) {
        AFImageDownloadReceipt *receipt = [self.downloader
                                           downloadImageForURLRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:imageURLString]]
                                           success:nil
                                           failure:nil];
        [self.downloader cancelTaskForImageDownloadReceipt:receipt];
    }

    XCTestExpectation *expectation = [self expectationWithDescription:[NSString stringWithFormat:@"Image download should be cancelled immediately"]];
    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(3 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
        for (NSString *imageURLString in imageURLStrings) {

            // The cache image is still exist, but this cache image should be nil.
            UIImage *cacheImage = [self.downloader.imageCache imageWithIdentifier:imageURLString];
            XCTAssertNil(cacheImage);
        }
        [expectation fulfill];
    });

    [self waitForExpectationsWithCommonTimeout];
}

**Comment by kinarobin** [comment]
@jshier  This just same as [AlamofireImage #368](https://github.com/Alamofire/AlamofireImage/pull/368)

File: Tests/Tests/AFImageDownloaderTests.m [hunk 1]
```diff
@@ -512,6 +512,37 @@ - (void)testThatItCanDownloadAndCancelAndDownloadAgain {
...
+    
+    for (NSString *imageURLString in imageURLStrings) {
+        AFImageDownloadReceipt *receipt = [self.downloader
+                                           downloadImageForURLRequest:[NSURLRequest requestWithURL:[NSURL URLWithString:imageURLString]]
+                                           success:nil
+                                           failure:nil];
+        [self.downloader cancelTaskForImageDownloadReceipt:receipt];
+    }
+    
+    XCTestExpectation *expectation = [self expectationWithDescription:[NSString stringWithFormat:@"Image download should be cancelled immediately"]];
```
jshier: Since there's no format, using `stringWithFormat` is unnecessary.

Tests/Tests/AFImageDownloaderTests.m
```diff
@@ -512,6 +512,37 @@ - (void)testThatItCanDownloadAndCancelAndDownloadAgain {
...
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(3 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
jshier: A 3 second delay seems overly long, given the downloads should be immediately cancelled. I think moving it down to a 1 second delay should be enough, but it would be best if there was a way to only assert once the downloads are completed in their cancelled state.

Tests/Tests/AFImageDownloaderTests.m
kinarobin: Thanks.

Tests/Tests/AFImageDownloaderTests.m
kinarobin: Done!

PR #4409, opened 2019-06-25
Fix NSSecureCoding

ElfSundae:
Using`decodeObjectOfClass:[NSNumber class]` to decode an NSNumber object previously encoded by `encodeInteger` will fail.


virusman:
@jshier Is a release planned with this fix?

PR #4418, opened 2019-07-25
Fix build error when using UIKit for Mac (Project Catalyst)

revolter:
As the error states:

> 'initWithMemoryCapacity:diskCapacity:diskPath:' is unavailable: not
  available on macCatalyst

and it will also be deprecated in iOS 13.


itayAmza:
When will it be committed to Master? is there any official support for Catalyst expected soon?

andytriboletti:
I would also like to see this work for Catalyst apps. Thanks for your efforts.

jshier:
Thanks, but this will be handled in a separate PR.

PR #4423, opened 2019-08-09
change api and add available url

Santiago0412:
The post API is deprecated and request is not Availble, we can use this http://myjson.com to add our own json test API, here is a simple example in my code.


jshier:
Unfortunately, we can't have our example app run off an API that can be publicly changed, or is controlled outside of our organization. If you'd like to find another API that can serve as a replacement, feel free to open another PR.

PR #4424, opened 2019-08-10
Added didReceiveDataBlock, didReceiveResponseBlock to AFURLSessionManager

leshkoapps:
Added didReceiveDataBlock, didReceiveResponseBlock to AFURLSessionManager. These blocks are useful when you need small portion of data received from the server but not all data in the completion block. I use these blocks in my audio streaming apps.


AFNetworking/AFURLSessionManager.m
```diff
@@ -290,7 +303,12 @@ - (void)URLSession:(__unused NSURLSession *)session
     self.downloadProgress.totalUnitCount = dataTask.countOfBytesExpectedToReceive;
     self.downloadProgress.completedUnitCount = dataTask.countOfBytesReceived;
 
-    [self.mutableData appendData:data];
+    if(self.didReceiveDataBlock){
```
jshier:
```suggestion
    if(self.didReceiveDataBlock) {
```

AFNetworking/AFURLSessionManager.m
```diff
@@ -621,7 +639,7 @@ - (void)setDelegate:(AFURLSessionManagerTaskDelegate *)delegate
     [self.lock unlock];
 }
 
-- (void)addDelegateForDataTask:(NSURLSessionDataTask *)dataTask
+- (AFURLSessionManagerTaskDelegate *)addDelegateForDataTask:(NSURLSessionDataTask *)dataTask
```
jshier: Please adjust the alignment of all of these method blocks so the colons line up.

AFNetworking/AFURLSessionManager.h
```diff
@@ -232,6 +232,11 @@ NS_ASSUME_NONNULL_BEGIN
                              downloadProgress:(nullable void (^)(NSProgress *downloadProgress))downloadProgressBlock
                             completionHandler:(nullable void (^)(NSURLResponse *response, id _Nullable responseObject,  NSError * _Nullable error))completionHandler;
 
+- (NSURLSessionDataTask *)dataTaskWithRequest:(NSURLRequest *)request
```
jshier: It seems like this method should take all of the previous parameters as well as the new blocks while the older methods internally call this one with nil blocks, rather than bifurcating the API, both publicly and internally.

jshier:
@leshkoapps Thanks for your work here. However, due to time constraints I won't be able to merge this for the 4.0 release, given its rather significant change to core library behaviors. So thank you again for your work here, and if we open the library back up to new features in the future, feel free to submit the changes again.

PR #4432, opened 2019-09-07
sanitizing headers (maybe behind flag option?)

huguesbr:
Nginx will silently remove any headers containing `_` in the header field name. 
Maybe AFNetworking should sanitized any header (maybe behind a flag option, or at least warn?)

https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#missing-disappearing-http-headers


jshier:
Thanks for the PR, but we won't be merging it. Thanks anyway!

huguesbr:
@jshier haha I guess new year PR clean up..

PR #4443, opened 2019-09-24
fix(location AFURLRequestSerialization.m): Fix implicit conversion ch…

DwightDing:
Fix implicit conversion changes signedness when build with Xcode 11.0


jshier:
Thanks for the PR, but this has been fixed by #4409.

PR #4439, opened 2019-09-16
Feature/wk web view

tjanela:
This PR leverages the previous #4430 from @Tripwire999

We have updated the httpbin.org certificate and its trust chain certificates.
We have updated the ```[AFWKWebView testProgressIsSet]``` test to use a large download URL.

PR #4464, opened 2019-10-22
Fix travis CI build stall on osx image xcode7.3

tjanela:
@jshier, this is the second in a series of PRs that aim to breakdown #4439 into smaller chunks.
This one updates the travis yaml file to use a pinned version of fastlane. 
The latest versions of fastlane caused a build stall on osx image xcode7.3.

We used the last successful travis build to identify what version of fastlane was used there and pinned fastlane to that version. 
It is unknown if a more recent (but not more recent than 2.127.1) version causes the same behavior. 
We do know that versions released after, and including, 2.127.1 cause this behavior.


jshier:
This doesn't seem to actually fix anything. It looks like those Travis images are far out of date, so why don't we go ahead and just remove them instead. We can support Xcode 8+.

bofeizhu:
@tjanela @jshier I opened a new PR to update CI settings. #4472

jshier:
This PR is no longer necessary, but thanks anyway!

PR #4519, opened 2020-03-16
feat: rerquest merge

wangjufan:
1. Merge multiple requests of the some data item  to one but return one by one .
2. Merge a list of different data item requests to one but return one by one  .
3. Automatic notify the receiptor when data item was refreshed one by one .
It's an excellent solution to reduce the complexity of response for request with an id .


jshier:
Sorry, we aren't taking any new features.

PR #4522, opened 2020-03-27
Fix compile regression, warnings, remove availability, prepare for 4.0.0 release

Kaspik:
- fixes regression from https://github.com/AFNetworking/AFNetworking/pull/4007 - missing headers in header file
- removes unused availability as it's now supported iOS10+, macOS 10.10+, watchOS 2+ and tvOS 10+
- fixes warning with CoreServices - https://github.com/AFNetworking/AFNetworking/issues/4520
- drops iOS 8, iOS 9 and macOS 10.9 (macOS 10.9 doesn't support WKWebView, it makes no sense to keep support of this version, and iOS versions were kind of messy in the project, via screenshot below)
- fixes all warnings in project for XCode 11.4

Screen Shot 2020-03-27 at 14 00 48

PR #4523, opened 2020-03-27
AFNetworking 4

jshier:
This PR tracks the changes for AFNetworking 4.

Issues remaining:
- [x] Failing encoding test on Catalyst.
- [x] Failing trust tests on Catalyst.

These failures are strange, all macOS and iOS tests pass.


Kaspik:
Please also update the podspec to fix warnings for projects - https://github.com/AFNetworking/AFNetworking/pull/4522/files#diff-1f8ef09d890801d130cd1ee773e1cb2eR38

jshier:
@Kaspik How would I make those includes conditional for all deployment targets?

Kaspik:
They are iOS / watchOS only. They were renamed, but it's still the same SDK, so projects will handle that correctly afaik.

jshier:
Right, but the rename is only valid when built using the 13.4 SDK right?

Kaspik:
Tests here should catch that because they are not using XCode 11.4 (or if you have one), or we could create dummy project for that - but I hope it's backwards compatible, I don't want to run another branch for years to silence the warning. 😭

jshier:
There is no warning in the project locally, likely due to the fact that our deployment targets are lower. If the rename works with Xcode 11+ and doesn't require 11.4, then I'm fine doing the rename. I'll try it out locally and see.

jshier:
Building using Xcode 11.3 fails if we only import `CoreServices.h`.

Kaspik:
Allright, let's release 4.0.0 and then we can separate branch/PR/release after that with this change. Thanks for testing!

jshier:
I do wonder if there's some condition we can check to avoid importing MobileCoreService on newer versions. I'll look into it.

bofeizhu:
Hi @jshier, it looks like Xcode 11.4 is not yet available on Travis.

jshier:
@bofeizhu Thanks. I'm not sure that's my only problem but it certainly doesn't help.

Kaspik:
Seeing all the Travis problems, did you consider using Github Actions?

jshier:
@Kaspik Travis wasn't really the problem, more my confusion around their environment and the old Fastlane stuff that this library uses. However, that does remind me that Travis doesn't yet support 10.15 while GitHub does, so switching would allow us to test Catalyst. And I don't think it's a problem that they only support Xcode 11, as Apple will start requiring that in June (was April) anyway.

Kaspik:
Cool! Checkout MessageKit setup I did for Github Actions if you want - https://github.com/MessageKit/MessageKit/pull/1297

jshier:
There we go. And much faster too.

Kaspik:
Yeah Github Actions are pretty awesome, yet expensive for macOS private repos.

jshier:
@kinarobin @ElfSundae @Kaspik Please take another look on master for any last minute fixes or changes you'd like to see. I'm preparing the 4.0 release now.

Package.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import PackageDescription
+
+let package = Package(name: "AFNetworking",
+                      platforms: [.macOS(.v10_10),
+                                  .iOS(.v9),
+                                  .tvOS(.v9),
+                                  .watchOS(.v2)],
+                      products: [.library(name: "AFNetworking",
```
Kaspik: Was there any reason why AFNetworking is not enabled to be dynamic library?

`.library(name: "AFNetworking",, type: .dynamic, targets: ["AFNetworking"])`

cc @joshfriend

jshier: That setting should be controlled by consumers, not the library.

Kaspik: Yeah, if consumer uses Package.swift and adds AFNetworking as a dependency, then it's possible.

But projects using SPM through XCode don't have this option yet and it gets imported always as a static library... 😞 
https://forums.swift.org/t/how-to-link-a-swift-package-as-dynamic/32062

Screenshot 2020-03-30 at 00 43 28


Thanks tho.

jshier: Yeah, unfortunately Xcode integrations are always static, no matter what the library says.

PR #4530, opened 2020-03-30
Unify bundle identifier for watchOS framework

ElfSundae:
Now the bundle identifiers of frameworks on all platforms are the same `com.alamofire.AFNetworking`. You may check this in the Build/Products directory, or in Carthage output.


jshier:
We must have a separate identifier for watchOS, as we found [on Alamofire](https://github.com/Alamofire/Alamofire/issues/2925). It's an iTMS requirement.

ElfSundae:
@jshier 👍 Thank you so much for mentioning that. I never knew.

PR #4536, opened 2020-03-31
Fix UIKit+AFNetworking.h

ElfSundae:
For manual installation, adding `|| TARGET_OS_TV` conditional to expose some categories to tvOS.

### DIFF
--- a/UIKit+AFNetworking/UIKit+AFNetworking.h
+++ b/UIKit+AFNetworking/UIKit+AFNetworking.h
@@ -20,23 +20,24 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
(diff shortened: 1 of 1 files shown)


layeng:
Hi @ElfSundae, I got this problem 2 days ago and finally figured out what the comments and your solution meant. However, it means having to overwrite  UIKit+AFNetworking.h which comes first as read only. I got it when I did a fresh new pod install. Do you know of any way of doing pod install with the UIWebView replaced by WKWebView? That is the problem is removed? Thank you and looking forward to your reply.

ElfSundae:
@layeng I don't understand your question very clearly. Do you want to use CocoaPods to install AFNetworking without UIWebView category? AFNetworking 4.x has already replaced UIWebView with WKWebView, you just need to update your dependency, i.e. `$ pod update AFNetworking`.  See [README](https://github.com/AFNetworking/AFNetworking#installation-with-cocoapods) for the installation.

Issue #4541, opened 2020-04-03
Crashed!


KorolZhu:
![企业微信截图_d94dbd4a-d384-4388-bb95-69a82e534254](https://user-images.githubusercontent.com/1926639/78331498-bc314800-75b8-11ea-8dd4-10f44a17b60a.png)

 [__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[2]' 

serverTrust is nil

Kaspik:
Following, new crash on latest AFNetworking 4.0.0.

@jshier Worth fixing and releasing 4.0.1 with all the changes on master?

Kaspik:
Still an issue waiting for merge.

Kaspik:
Can be closed.

jshier:
Fix released in 4.0.1.

PR #4542, opened 2020-04-03
Fix serverTrust crash

Kaspik:
### Issue Link :link:
- Fixes https://github.com/AFNetworking/AFNetworking/issues/4541

### Goals :soccer:
- Release of the new version 4.0.1 with no crash.

`challenge.protectionSpace.serverTrust` is nil if the authentication method of the protection space is not server trust


0xced:
Thanks for the fix! Can you add a unit test that triggers the crash (without your fix) in order to avoid a future potential regression?

0xced:
@Kaspik May I ask why did you close this pull request?

Kaspik:
Hah, I merged my other PR and it closed this one. I'll reopen.

0xced:
Are you comfortable to add a unit test that triggers the crash (without your fix) in order to avoid a future potential regression or would you prefer me to add the unit test?

Kaspik:
I'm already out from my computer for the weekend, so feel free to do it otherwise I'll try on Monday. :)

3a4oT:
ping @Kaspik

Kaspik:
As `static NSError * ServerTrustError(SecTrustRef serverTrust, NSURL *url)` was private, I ended up updating it to func that is publicly exposed, then adding test it throws if serverTrust is nil.

Let me know if you have any other idea how to revert this but still test it.

AFNetworking/AFURLSessionManager.h
```diff
@@ -447,6 +447,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)setDownloadTaskDidResumeBlock:(nullable void (^)(NSURLSession *session, NSURLSessionDownloadTask *downloadTask, int64_t fileOffset, int64_t expectedTotalBytes))block;
 
++ (nonnull NSError *)serverTrustErrorWithServerTrust:(nonnull SecTrustRef)serverTrust url:(nonnull NSURL *)url;
```
jshier, line 450: I'd rather not make this method public. Instead of testing the error method directly, perhaps you can test the delegate method which creates the error? I do like the nullability annotations though, so applying those to the function might be helpful.

Kaspik: I'm not sure how to "repro" the case where it returns with no server trust - so if anyone has, feel free to give me a hint or final solution.

jshier:
Merging to branch to add tests but will credit this PR with the change.

PR #4550, opened 2020-04-18
Replace instead of appending to the default header

ElfSundae:
Related issue: https://github.com/AFNetworking/AFNetworking/issues/4549
For configuring headers to an individual request, in general, we prefer to replace the default value, instead of appending things to the default value with a comma, right?
In my opinion, using `setValue` is more reasonable and clearer.

### DIFF
--- a/AFNetworking/AFHTTPSessionManager.m
+++ b/AFNetworking/AFHTTPSessionManager.m
@@ -180,7 +180,7 @@
     NSError *serializationError = nil;
     NSMutableURLRequest *request = [self.requestSerializer multipartFormRequestWithMethod:@"POST" URLString:[[NSURL URLWithString:URLString relativeToURL:self.baseURL] absoluteString] parameters:parameters constructingBodyWithBlock:block error:&serializationError];
     for (NSString *headerField in headers.keyEnumerator) {
-        [request addValue:headers[headerField] forHTTPHeaderField:headerField];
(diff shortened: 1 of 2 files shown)


ElfSundae:
Setting headers per request is introduced from #4113 . At first I also doubted whether there were special reasons to use `addValue`, I searched the history and no other discussion about this.
Yes, tests are neccessary, but may be later.

jshier:
Sorry, tests need to be added as part of this PR, and really any code change going forward. It's the only way to ensure the integrity of the library.

Also, would you mind summarizing the issue this fixes?

ElfSundae:
Tests are added. "later" I meant is now. 😄 I was going out.

ElfSundae:
> would you mind summarizing the issue this fixes?

Do you mean updating the changelog file? Sorry I'm afraid I can't cause I'm not good at English :(

ElfSundae:
It might be:

**Changes**

- Setting HTTP headers per request in `AFHTTPSessionManager` methods now replaces the default value specified in the request serializer.

jshier:
> > would you mind summarizing the issue this fixes?
> 
> Do you mean updating the changelog file? Sorry I'm afraid I can't cause I'm not good at English :(

No, I just meant that you should fill out the PR template so that everyone can see why you're making a change and how you did so. Otherwise there might be a lot of unnecessary questions during review.

ElfSundae:
PR description was updated.

PR #4551, opened 2020-04-18
Fix nullability

ElfSundae:


AFNetworking/AFURLRequestSerialization.h
```diff
@@ -216,7 +216,7 @@ forHTTPHeaderField:(NSString *)field;
 
  @param block A block that defines a process of encoding parameters into a query string. This block returns the query string and takes three arguments: the request, the parameters to encode, and the error that occurred when attempting to encode parameters for the given request.
  */
-- (void)setQueryStringSerializationWithBlock:(nullable NSString * (^)(NSURLRequest *request, id parameters, NSError * __autoreleasing *error))block;
+- (void)setQueryStringSerializationWithBlock:(nullable NSString * _Nullable (^)(NSURLRequest *request, id parameters, NSError * __autoreleasing *error))block;
```
jshier, line 219: This seems redundant: the `NSString` return value is already nullable, this is just the same annotation spelled another way.

AFNetworking/AFURLRequestSerialization.h
```diff
@@ -234,10 +234,10 @@ forHTTPHeaderField:(NSString *)field;
 
  @return An `NSMutableURLRequest` object.
  */
-- (NSMutableURLRequest *)requestWithMethod:(NSString *)method
-                                 URLString:(NSString *)URLString
-                                parameters:(nullable id)parameters
-                                     error:(NSError * _Nullable __autoreleasing *)error;
+- (nullable NSMutableURLRequest *)requestWithMethod:(NSString *)method
```
jshier, line 237: This doesn't seem necessary, as there's no way the method can return `nil`. What path does it take to do so for you?

AFNetworking/AFHTTPSessionManager.h
```diff
@@ -271,14 +271,14 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
-- (NSURLSessionDataTask *)dataTaskWithHTTPMethod:(NSString *)method
-                                       URLString:(NSString *)URLString
-                                      parameters:(nullable id)parameters
-                                         headers:(nullable NSDictionary <NSString *, NSString *> *)headers
-                                  uploadProgress:(nullable void (^)(NSProgress *uploadProgress)) uploadProgress
-                                downloadProgress:(nullable void (^)(NSProgress *downloadProgress)) downloadProgress
-                                         success:(nullable void (^)(NSURLSessionDataTask *task, id _Nullable responseObject))success
-                                         failure:(nullable void (^)(NSURLSessionDataTask * _Nullable task, NSError *error))failure;
+- (nullable NSURLSessionDataTask *)dataTaskWithHTTPMethod:(NSString *)method
```
jshier, line 274: I know this matches the other methods, but it doesn't make much sense these are all `nullable` return values.

jshier, line 274: Could you ensure the implementations have matching nullability as well?

AFNetworking/AFURLRequestSerialization.h
ElfSundae: `nullable` is for the method parameter, so you can pass `nil` to this method: `setWithBlock:nil`.
`_Nullable` is for the return value of the block, so you can return `nil` in the block if any serialization error occurs:
```
setWithBlock:^NSString * _Nullable (...) {
    *error = [NSError errorWith...];
    return nil;
}
```

jshier: No, the [block syntax](http://fuckingblocksyntax.com) is as follows: `(returnType (^nullability)(parameterTypes))blockName;`. So the code currently allows for returning `nil` just fine, it just doesn't allow for passing a `nil` block. Why would we want to allow a `nil` block parameter here?

AFNetworking/AFURLRequestSerialization.h
ElfSundae: Inside this method, it calls `-requestBySerializingRequest:withParameters:error:` which declares `nullable` return value in `AFURLRequestSerialization` protocol as serializing query parameters may be failed.

AFNetworking/AFHTTPSessionManager.h
ElfSundae: The same reason: because of the `serializationError`. It may return `nil` indeed:

```
if (serializationError) {
    if (failure) {
        dispatch_async(self.completionQueue ?: dispatch_get_main_queue(), ^{
            failure(nil, serializationError);
        });
    }

    return nil;
}
```

AFNetworking/AFURLRequestSerialization.h
jshier: Ah, makes sense.

AFNetworking/AFHTTPSessionManager.h
jshier: Thanks.

AFNetworking/AFURLRequestSerialization.h
ElfSundae: I think you may have made a mistake.

Yes, `nullability returnType (^) ...` is equivalent to `returnType (^nullability) ...` . But this `nullability` is a qualifier for the whole block, and the block is a parameter of the `setQueryStringSerializationWithBlock:` method. That means this method accepts one parameter and the parameter can be `nullability` (`nil` is this case).

> The code currently allows for returning nil

Not **returning** nil, but **passing** nil.

Users may pass a block to customize the query string serialization, or pass `nil` to reset it to use the default serialization provided by `AFURLRequestSerialization`.

I added `_Nullable` for the return type of the block, because the block can return `nil` if any error occurs.

You may test current return type of the block via Xcode Autocompletion:

```objc
[[AFHTTPRequestSerializer serializer]
    setQueryStringSerializationWithBlock:^NSString * _Nonnull(NSURLRequest * _Nonnull request, id  _Nonnull parameters, NSError *__autoreleasing  _Nullable * _Nullable error) {

}];
```

You may check out [`setDownloadTaskDidFinishDownloadingBlock`](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFURLSessionManager.h#L434) it also follows this style `nullable blockReturnType _Nullable (^)...`.

AFNetworking/AFHTTPSessionManager.h
ElfSundae: I checked all `dataTaskWith...`, `requestWith...` and RESTful methods in the session manager and the request serializer. The others have correct nullability except for these two methods.

In addition, accurately, the multipart `POST` method will never return `nil`. The multipart `POST` method invokes `[self.requestSerializer multipartFormRequestWithMethod:...]` which returns `nonnull` becuase it passes `nil` as `parameters` to `-requestWithMethod:URLString:parameters:error` that won't cause an `serializationError`.

Then we have three options:

Option 1: add `nullable` to the return type for `-multipartFormRequestWithMethod`, to match other request generator methods. And check the return value from `requestWithMethod`:

```diff
  NSMutableURLRequest *mutableRequest = [self requestWithMethod:method ..

+ if (!mutableRequest) {
+    return nil;
+ }

  // append multiparts

```

Option 2: remove `nullable` from the multipart `POST` method. And remove the `serializationError` checker?

```diff
- NSError *serializationError = nil;
  NSMutableURLRequest *request = [self.requestSerializer multipartFormRequestWithMethod... 
+                                   error:NULL];
- if (serializationError) {
-    if (failure) {
-         dispatch_async(self.completionQueue ?: dispatch_get_main_queue(), ^{
-            failure(nil, serializationError);
-        });
-    }
-
-     return nil;
- }

 // create task
```

Option 3: change nothing :)

This PR chooses the third option, now I prefer option 1, what do you think?

AFNetworking/AFURLRequestSerialization.h
jshier: It appears you're correct. The first `nullable` controls the nullability of the block, the `_Nullable` controls the nullability of the return value.

AFNetworking/AFHTTPSessionManager.h
jshier: That seems fine for now, it can be updated if necessary in the future.

AFNetworking/AFURLRequestSerialization.h
Kaspik: This is btw breaking change for Swift - as from previous method without optional return value now it's a method that throws.

Screen Shot 2020-04-22 at 20 34 12

jshier: That's probably a good thing, as it means the importer wasn't properly transforming the error parameter into a `throw`. We should audit the other APIs to find similar cases, as I'd consider them bugs in the API visible to Swift.

PR #4552, opened 2020-04-18
Fix condition of evaluateServerTrust

### LINKED ISSUE #4541: Crashed!

ElfSundae:
Related to #4542, fixes #4541

We should evaluate server trust only when performing **server trust authentication**.


jshier:
This condition is more correct that the other, so I'm combining the two and both will get credit.

PR #4553, opened 2020-04-19
Fix ServerTrustError crash.

jshier:
### Issue Link :link:
#4541

### Goals :soccer:
This PR builds on #4542 and #4552 to fix the crash seen when attempting to create an error for a `nil` `SecTrustRef`. It combines both approaches:

- Ensure we only attempt to check the trusts when `[challenge.protectionSpace.authenticationMethod isEqualToString:NSURLAuthenticationMethodServerTrust]` is `YES`.
- When creating an error, ensure proper nullability checks when inserting values into the `userInfo` dictionary, as even in the `NSURLAuthenticationMethodServerTrust` case, the values can be `nil`.

This PR makes the previous 

```objc
static NSError * ServerTrustError(SecTrustRef serverTrust, NSURL *url)
```
 convenience function into an internal method on `AFURLSessionManager`

```objc
- (nonnull NSError *)serverTrustErrorForServerTrust:(SecTrustRef)serverTrust url:(NSURL *)url`
```

to enable testing and adds additional checking internally to ensure no `nil` values are inserted into the dictionary.

### Testing Details :mag:
Tests were added for each permutation of parameters to the new method. To enable this, a `SecTrust` creation function was moved into `AFTestCase` to enable its use in other tests.

### DIFF
--- a/Tests/Tests/AFHTTPSessionManagerTests.m
+++ b/Tests/Tests/AFHTTPSessionManagerTests.m
@@ -28,6 +28,12 @@
 @property (readwrite, nonatomic, strong) AFHTTPSessionManager *sessionManager;
 @end
 
+@interface AFURLSessionManager (Testing)
+
+- (nonnull NSError *)serverTrustErrorForServerTrust:(SecTrustRef)serverTrust url:(NSURL *)url;
+
+@end
+
 @implementation AFHTTPSessionManagerTests
 
 - (void)setUp {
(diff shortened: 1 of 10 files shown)


jshier:
@0xced Since you commented on the other PR, would you mind taking a look?

@Kaspik @ElfSundae This combines your previous PRs. Please take a look when you get a chance to ensure it matches your expectation.

Tests/Tests/AFHTTPSessionManagerTests.m
```diff
@@ -28,6 +28,12 @@ @interface AFHTTPSessionManagerTests : AFTestCase
 @property (readwrite, nonatomic, strong) AFHTTPSessionManager *sessionManager;
 @end
 
+@interface AFURLSessionManager (Testing)
+
+- (nonnull NSError *)serverTrustErrorForServerTrust:(SecTrustRef)serverTrust url:(NSURL *)url;
```
Kaspik, line 33: ServerTrust and url could have nullability too. :)

ElfSundae:
I think this may be a bit over-processing.

- `serverTrust` may be `nil` even in the server trust authentication? I have not met this issue yet.

  According to the [Apple doc](https://developer.apple.com/documentation/foundation/nsurlprotectionspace/1409926-servertrust?language=objc):

  > This value is `nil` if the authentication method of the protection space is not server trust.

- How the task URL is `nil` inside `didReceiveChallenge` callback?

- minor: `nonnull` could be removed to follow our style :)

jshier:
Unfortunately those are not API level guarantees, so they can't form the basis of our implementation. Both the `serverTrust` and the `task.currentRequest` are `nullable` in the API, so we have to treat them as such at every point.

I'll double check the nullability.

jshier:
Doesn't look like `NS_ASSUME_NONNULL` is active in the `.m` file, so `nullable` is the default, and we have to mark `nonnull` explicitly.

jshier:
Merging this for release, but still feel free to review, @0xced.

PR #4554, opened 2020-04-19
Fix SPM Usage

jshier:
### Issue Link :link:
Original Package.swift introduced in 4.0.0 didn't work due to an improper header search path.

### Goals :soccer:
This PR fixes the Package.swift file by setting the `publicHeadersPath` value correctly.

As explained [here](https://forums.swift.org/t/can-i-use-swift-package-manager-in-obj-c-code/35463/16), the `publicHeadersPath` value is relative to the target directory, not the package file, so it should've been `""` instead of `"AFNetworking"`.

### Testing Details :mag:
Tested manually with a test project that was broken previously.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -34,4 +34,4 @@ let package = Package(name: "AFNetworking",
                                           targets: ["AFNetworking"])],
                       targets: [.target(name: "AFNetworking",
                                         path: "AFNetworking",
-                                        publicHeadersPath: "AFNetworking")])
+                                        publicHeadersPath: "")])


zarghol:
Hi ! I think this leads a problem in our side :
```sh
/Users/XXXX/Library/Developer/Xcode/DerivedData/YYYY/Build/Intermediates.noindex/ArchiveIntermediates/ZZZZ/IntermediateBuildFilesPath/GeneratedModuleMaps/iphoneos/AFNetworking.modulemap:2:10: umbrella directory '/Users/XXXX/Library/Developer/Xcode/DerivedData/AAAAAA/SourcePackages/checkouts/AFNetworking/AFNetworking/AFNetworking' not found
```
I don't know the specification of public header path in SPM, but maybe it searches it at the root with a string empty, but it is inside the folder AFNetworking, not at the root...

PR #4557, opened 2020-04-21
ignore https check if baseURL is not set

mojtabacazi:
### Goals :soccer:
We should allow usage of security policy if URL construction is being managed outside of AFNetworking and not using `baseURL` variable.

Check usage of `https` only if `baseURL` is set. Otherwise assume it's user responsibility to use the right protocol.

### Testing Details :mag:
Tests are updated to validate new changes.


jshier:
Thanks for the PR!

I'm inclined to lift this restriction, but this PR doesn't go far enough in updating the logic. Please outline the conditions in which we should produce an error when applying a security policy and update the tests accordingly.

PR #4565, opened 2020-05-07
Use the latest patch version in README

matthewmayer:
### Goals :soccer:
Make README up to date

Specifying 4.0 instead of full patch version caused problem e.g. as at https://stackoverflow.com/questions/61655457/cocoapods-could-not-find-compatible-versions-for-pod-afnetworking/61657341#61657341

### Testing Details :mag:
Documentation update only


jshier:
Thanks for the PR! However, these are correct: all of the package managers should properly update to the latest version with the current version strings, it's usually just a matter of "update" vs. "install" actions. I like to avoid having to update the README for every version bump. Thanks anyway!

Issue #4570, opened 2020-05-13
AFNetworkReachabilityManager's property "reachable" not accurate.


macro-c:
ℹ when i call AFNetworkReachabilityManager's property "reachable" to find out the current network status on my device, it will return NO, when that happend, my device's wifi and Cellular network’switch is on; 
then I find that property "networkReachabilityStatus" was init with value "AFNetworkReachabilityStatusUnknown", after a few mement it refresh to "AFNetworkReachabilityStatusReachableViaWiFi" by a block；

## What did you expect to happen?

ℹ in method 
`- (instancetype)initWithReachability:(SCNetworkReachabilityRef)reachability`,  
use the code below instead of init networkReachabilityStatus with AFNetworkReachabilityStatusUnknown;
    `SCNetworkReachabilityFlags flags;`
    `if (SCNetworkReachabilityGetFlags(self.networkReachability, &flags)) {`
            `self.networkReachabilityStatus = AFNetworkReachabilityStatusForFlags(flags);`
     `} else {`
             `self.networkReachabilityStatus = AFNetworkReachabilityStatusUnknown;`
     `}` 

if there is other concern about this issue, please let me know.

thanks

ElfSundae:
You need to call `-startMonitoring` first.

macro-c:
YES，i have call startMonitoring before use property networkReachabilityStatus.
i find that the block where refresh property networkReachabilityStatus is called after a few moment.

ElfSundae:
Yes, it needs time to retrieve reachability status because of `SCNetworkReachabilityGetFlags`.
FYI https://developer.apple.com/library/archive/qa/qa1693/_index.html
> The System Configuration framework reachability API (<SystemConfiguration/SCNetworkReachability.h>) operates synchronously by default. Thus, seemingly innocuous routines like `SCNetworkReachabilityGetFlags` can get you killed by the watchdog. If you're using the reachability API, you should use it asynchronously.

macro-c:
thanks, now i know why do not use `SCNetworkReachabilityGetFlags`;

under these circumstances, the method `- (BOOL)isReachable`  should be 
`self.networkReachabilityStatus != AFNetworkReachabilityStatusNotReachable` 
rather than `[self isReachableViaWWAN] || [self isReachableViaWiFi]`

ElfSundae:
I don't think so, The current `-isReachable` is right and clear enough - returns `YES` only when the device is reachable.
[...]

PR #4571, opened 2020-05-13
networkReachabilityStatus init with a right status

macro-c:
### Issue Link :
https://github.com/AFNetworking/AFNetworking/issues/4570

### Testing Details :
fix this issue


jshier:
If there is an issue here, this wouldn't be the correct approach, as calculating the reachability state should not be done synchronously in the initializer. The issue seems to have reach a conclusion so I'm closing this PR.

PR #4603, opened 2020-07-20
simplified code part of WKWebView+AFNetworking

Dcell:
### Issue Link :link:
simplified code part of WKWebView+AFNetworking

1.remove `__block` before `*dataTask`
2.modify baseURL '[dataTask.currentRequest URL]' -> `[request URL]`

### Testing Details :mag:
AFWKWebViewTests [PASS]


jshier:
Thanks for the PR! However, we won't be merging it as it doesn't appear to fix or enhance any behavior, and we'd like to avoid changes that aren't absolutely necessary.

PR #4614, opened 2020-09-11
Correct AFURLSessionManager comments.

Huang-Libo:


jshier:
Looks correct, thanks.

Issue #4616, opened 2020-09-17
Add UIKit extensions in swift package manager


tc-sgupta:
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/AFNetworking/AFNetworking/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag `afnetworking`.
>
> Please remove this line and everything above it before submitting.

* [ ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/AFNetworking/AFNetworking/blob/master/CONTRIBUTING.md).

## What did you do?

ℹ Please replace this with what you did.  

## What did you expect to happen?

ℹ Please replace this with what you expected to happen.  

## What happened instead?

ℹ Please replace this with of what happened instead.  

## AFNetworking Environment

**AFNetworking version:**
**Xcode version:**
**Swift version:**
**Platform(s) running AFNetworking:**
**macOS version running Xcode:**

## Demo Project

ℹ Please link to or upload a project we can download that reproduces the issue.

rodmaz:
Same need here.

mkowalski87:
👍

JuhaO81:
Need this also :)

jshier:
Feel free to put up a PR.

tc-sgupta:
> Feel free to put up a PR.

@jshier I will create a PR

tc-sgupta:
@jshier Please review this PR #4623

tc-sgupta:
@jshier Please have a look at the PR #4623

tc-sgupta:
@jshier Please have a look at the PR #4623

MapaX:
@jshier Any ETA when this could be merged?

JuhaO81:
Please merge and release :)

PR #4623, opened 2020-10-03
Added support to add UIKit+AFNetworking using Swift Package Manager

tc-sgupta:
### Issue Link :link:
https://github.com/AFNetworking/AFNetworking/issues/4616

### Goals :soccer:
- Added support to add UIKit+AFNetworking using Swift Package Manager

- Added target `UIKit_AFNetworking` depends on target `AFNetworking`

### Testing Details :mag:
- Create a project and try to add AFNetworking or UIKit_AFNetworking target.
Screen Shot 2020-10-04 at 12 05 17 AM

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,15 @@ let package = Package(name: "AFNetworking",
                                   .tvOS(.v9),
                                   .watchOS(.v2)],
                       products: [.library(name: "AFNetworking",
-                                          targets: ["AFNetworking"])],
+                                    targets: ["AFNetworking"]),
+                                .library(name: "UIKit_AFNetworking",
+                                    targets: ["UIKit_AFNetworking"])
+                                ],
                       targets: [.target(name: "AFNetworking",
                                         path: "AFNetworking",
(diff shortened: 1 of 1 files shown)


JuhaO81:
Any estimates when is this going in and to be released?

jshier:
Thanks for the PR! Unfortunately this makes the UIKit extensions available to all package consumers, not just those on iOS, so anyone who includes both packages and isn't on iOS will see a build failure. There doesn't seem to be a way to conditionalize the target so it's only available on iOS, which is why I didn't include it when creating the package in the first place. For that reason I'm inclined not to merge this PR and just say that anyone who wants the extensions should use other package managers, unless there's an alternate solution.

BenLeggiero:
As someone who consumes platform-dependent packages from multiple platforms, I know there are many mechanisms to remove the difficulties presented here. You could have a specific target just for macOS/AppKit or watchOS/WatchKit, or you could surround imports with compiler conditionals. Merging this will not break existing code, and those requiring it will be able to easily overcome any challenges they face including it.

PR #4629, opened 2020-10-30
Remove all references to UIWebView on 2.x version

Enricoza:
### Goals :soccer:
Remove all references to `UIWebView` to allow legacy apps that can't update to version 4.x to keep sending updates to the AppStore.

I'm aware that this is NOT an optimal fix. 
The optimal fix is the one released in the 4.x, which actually replaces `UIWebView` with `WKWebView`.
I'm also aware that this is a legacy branch that requires less attention from you, so implementing that same fix to this branch is way out of your scope.

That said: 
- given that Apple already stopped accepting new apps with any reference to `UIWebView`
- given that Apple will STOP accepting updates for apps with any reference to `UIWebView` within a month
- given that any app that is already using your `UIWebView` category will no longer be able to use it and successfully release an update to the AppStore

I think it's safe to just remove every reference to `UIWebView` from this branch to allow legacy apps to still send minor updates and bugfixes.

It may not be useful to many of your users, but I'm sure there is someone (like me) who has a really old app that:
- uses AFNetworking 2.x or 3.x and updating to 4.x would require too much work
- has dependencies that require AFNetworking 2.x or 3.x and don't receive any updates/are deprecated
- doesn't use the `UIWebView` category

I think that merging this pull request could help them a lot.

### Testing Details :mag:
I didn't test anything since this is just the removal of the `UIWebView` references.


ElfSundae:
This problem does exist, but I don't think it is a good idea to submit this breaking change to 2.x release.
For instance, I forked this repo and published an unofficial v3 CocoaPods library that replaced `UIWebView` with `WKWebView` (https://github.com/ElfSundae/AFNetworking/issues/1). For the users, they just need to add my source URL to the `Podfile`, then the `AFNetworking, ~> 3.0` dependency in either Podfile or podspec will be installed without `UIWebView`.

Enricoza:
Hi @ElfSundae,
You are right, this is a breaking change.. but as I said it SHOULD not affect anyone since nobody will be able to release any app to the AppStore with v2 anyway. But maybe I'm missing something, so don't just take my word for it.

Thank you anyway for your unofficial v3 release, that surely is helpful to anyone having the v3 at the moment.
Of course for the time being I'm using my own GitHub fork to remove the UIWebView references from my app. 
I just thought that it could be useful to have the fix on the main repo since it would be hard for anyone else to find my fork (or your fork) among the other 10k. 

Anyway if this can't be helpful or you guys think it's unsafe no problem... I solved my own issue anyway.

jshier:
Thanks for the PR @Enricoza, but we won't be accepting any additional changes to non-v4 AFNetworking versions, nor are we providing support for those versions except in the most extreme circumstances, like a security issue. I'd encourage anyone still on those versions to prioritize updating.

PR #4694, opened 2021-07-31
Fix 

leavesster:
### Goals :soccer:

when UIImageView download same URL twice, the second callback never get called. This behavior is different from the normal logic (normally, the first request will not call back, the second will call back when the request is not equal).

* different URL

```Objective-C
    [self.imageView
     setImageWithURLRequest:urlRequest1
     placeholderImage:nil
     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
        NSLog(@"first callback");// never be called
     }
     failure:nil];
    
    [self.imageView
     setImageWithURLRequest:urlRequest2
     placeholderImage:nil
     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
        NSLog(@"second callback");// will be called here
     }
     failure:nil];
```

* same URL

```Objective-C
    [self.imageView
     setImageWithURLRequest:self.jpegURLRequest
     placeholderImage:nil
     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
        NSLog(@"first callback");// will be called here
     }
     failure:nil];
    
    [self.imageView
     setImageWithURLRequest:self.jpegURLRequest
     placeholderImage:nil
     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
        NSLog(@"second callback");// never be called
     }
     failure:nil];
```

```Objective-C
- (void)setImageWithURLRequest:(NSURLRequest *)urlRequest
              placeholderImage:(UIImage *)placeholderImage
                       success:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, UIImage *image))success
                       failure:(void (^)(NSURLRequest *request, NSHTTPURLResponse * _Nullable response, NSError *error))failure
{
....

// cancel Image download when activeTaskURL is not equal to previous request
// but if activeTaskURL is equal to previous request, not return, just continue.
    if (![self isActiveTaskURLEqualToURLRequest:urlRequest]) {
        [self cancelImageDownloadTask];
    }

....
// cancel last same url request task, because you already add a new AFImageDownloaderResponseHandler to AFImageDownloaderMergedTask. here you cancel the previous task will not cancel the AFImageDownloaderMergedTask
        if ([self isActiveTaskURLEqualToURLRequest:urlRequest]) {
            [self cancelImageDownloadTask];
        }

        self.af_activeImageDownloadReceipt = receipt;
....
}
```

### Testing Details :mag:

[AFNetworking UIKit Tests] folder

AFUIImageViewTests.m

```Objective-C
- (void)testThatDuplicateRequestOnlyCalledLastTime
- (void)testThatDuplicateSameRequestOnlyCalledLastTime
```

### DIFF
--- a/Tests/Tests/AFUIImageViewTests.m
+++ b/Tests/Tests/AFUIImageViewTests.m
@@ -74,6 +74,51 @@
     [self waitForExpectationsWithCommonTimeout];
 }
 
+- (void)testThatDuplicateRequestOnlyCalledLastTime {
+    XCTAssertNil(self.imageView.image);
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Request should fail"];
+    [self.imageView
+     setImageWithURLRequest:self.jpegURLRequest
+     placeholderImage:nil
+     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
+        XCTFail("repeat requests should be called only the last time");
+     }
+     failure:nil];
+    
+    [self.imageView
+     setImageWithURLRequest:self.error404URLRequest
+     placeholderImage:nil
+     success:nil
+     failure:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, NSError * _Nonnull error) {
+        [expectation fulfill];
+    }];
+    [self waitForExpectationsWithCommonTimeout];
+}
+
+- (void)testThatDuplicateSameRequestOnlyCalledLastTime {
+    XCTAssertNil(self.imageView.image);
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Request should succeed"];
+    [self.imageView
+     setImageWithURLRequest:self.jpegURLRequest
+     placeholderImage:nil
+     success:^(NSURLRequest * _Nonnull request, NSHTTPURLResponse * _Nullable response, UIImage * _Nonnull image) {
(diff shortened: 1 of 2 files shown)


leavesster:
> Your code will cause repeated downloads. If the same request encapsulates a layer of judgment on the basis of AFN.

I don't think so. please show me the code if you have this example.

And even though this happened, this bug is also cause another problem. It is not an expected behavior

leavesster:
> No problem, do a `return` inside `downloadImageForURLRequest` 👍🏻.

can't understand what you say .

leavesster:
> I didn't look at the whole code before, AFN checked the existing tasks in this place, your code has no problem of repeated download.
> 
> ```
>         AFImageDownloaderMergedTask *existingMergedTask = self.mergedTasks[URLIdentifier];
>         if (existingMergedTask != nil) {
>             AFImageDownloaderResponseHandler *handler = [[AFImageDownloaderResponseHandler alloc] initWithUUID:receiptID success:success failure:failure];
>             [existingMergedTask addResponseHandler:handler];
>             task = existingMergedTask.task;
>             return;
>         }
> ```

I highly recommend you read AFN's whole code. I've learned a lot  from it.

iLiuChang:
👌🏻 thanks @leavesster

